PR #15, opened 2017-07-21
py3: print, line-endings

looooo:
there was a really strange bug with windows line endings. The message was:
```During initialization the error invalid character in identifier (<string>, line 1) occurred in /home/lo/.FreeCAD/Mod/FreeCAD_SheetMetal/InitGui.py```


looooo:
strange bug gets even stranger...

PR #16, opened 2017-07-21
py3 fixes

looooo:
this one should work now:

the problem was a strange starting character. No idea what it was used for. It isn't listed in the diff, but I think it is removed with the commits.


easyw:
I think for python != when None is present is suggested to use 'is not'
`if the_index != None:`
becomes
`if the_index is not None:`
and the same for i.e. 
`if the_index == None:`
becomes
`if the_index is None:`
Maurice

looooo:
`<>` is replaced with `!=`. So I would go with this syntax as it doesn't change the behavior.

shaise:
Thanks looooo, '<>' is indeed now obsolete. merging pull request.

looooo:
thanks

PR #36, opened 2018-06-26
adding Unfold exception message + using 'is' to check 'None'

easyw:
referring to this thread
https://forum.freecadweb.org/viewtopic.php?f=22&t=25281&p=241335#p241312
I'm adding an exception message

### DIFF
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -37,7 +37,7 @@ def smWarnDialog(msg):
     diag.exec_()
  
 def smBelongToBody(item, body):
-    if (body == None):
+    if (body is None):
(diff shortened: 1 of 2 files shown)


shaise:
thanks.

shaise:
whats the difference between 'is' and '==' in this regard?

easyw:
PEP 8 Style Guide
https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html
https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

shaise:
Thanks again!

PR #39, opened 2018-08-27
first step to py3 compatibility

easyw:
I'm testing the WB with this FC Py3Qt5 release:
0.18_pre/FreeCAD_0.18.14451_Conda_Py3QT5-WinVS2016_x64.7z
Still Unfolder needs more work...
(Py3 dict works differently)


shaise:
Thanks!
If dict works differently, will it be possible to make the same code work on both Py2 and Py3?

PR #46, opened 2018-09-18
OuterWire and Internal Sketches feature

easyw:
- creation of Outer sketch only when possible (outer wire is closed)
- option to select color for internal sketch added

### DIFF
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -2112,7 +2112,15 @@ except AttributeError:
     def _translate(context, text, disambig):
         return QtGui.QApplication.translate(context, text, disambig)
       
-      
+# loading presets
+pg = FreeCAD.ParamGet("User parameter:BaseApp/Preferences/Mod/sheetmetal")
+if pg.IsEmpty():
+    pg.SetBool("bendSketch",0)
+    pg.SetBool("genSketch",1)
+    pg.SetString("bendColor","#c00000")
+    pg.SetString("genColor","#000080")
+    pg.SetString("intColor","#ff5733")
+
 class QColorButton(QtGui.QPushButton):
     '''
     Custom Qt Widget to show a chosen color.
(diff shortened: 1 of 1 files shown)


shaise:
Hi easyw,

Looks good!
Two things we need to consider, though,
1. The unfold script is taken from Ulrich. He should be notified with your changes, or next versions will not be compatible. 
2. Why do you leave the "Unfold_sketch"? It is exactly same as "Unfold_sketch_OutWire" (BTW, I think Outline is better then OutWire)

shai

easyw:
> 1. The unfold script is taken from Ulrich. He should be notified with your changes, or next versions will not be compatible.

@ulrich1a please have a look at this PR (I also notified him at FC forum)
> 2. Why do you leave the "Unfold_sketch"? It is exactly same as "Unfold_sketch_OutWire"

This is not always true... in some case the outline wire is not closed (i.e. when the projection of the unfolded has extra edges on the cut or have internal edges not removed by the unfold script)
outline-sketch
wire-not-closed

> (BTW, I think Outline is better then OutWire)

I've updated it :smile:

shaise:
Thanks easyw.
I have merged the pull request + did some minor layout changes to the gui. 
I will investigate your code some more, perhaps I will use it to replace the current "unfold_sketch"

easyw:
@shaise 
> + did some minor layout changes to the gui.

I manually added the Gui code because the [ui file](https://github.com/shaise/FreeCAD_SheetMetal/blob/master/UnfoldOptions.ui) is outdated...

shaise:
yes I agree. Perhaps I need to remove it.

easyw:
It would be nice to have it updated... it is easier to contribute

shaise:
The problem is I dont know how to insert a custom widget in qdesigner

easyw:
an option is to create a simple button and then change its style
https://stackoverflow.com/questions/15080404/colouring-a-button-in-qt
Just save the ui and then add this code:
```
    mycolor = '#ff5733'
    ui.pushButtonColor.setStyleSheet("background: " + mycolor + ";");
    ui.pushButtonColor.setMaximumWidth(32)
```

[UnfoldOptions.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/2401599/UnfoldOptions.zip)

PR #48, opened 2018-10-28
Edge sel

jaisekjames:
* Added Support for selection based Edge.

### DIFF
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -106,7 +106,7 @@ def smRestrict(var, fromVal, toVal):
       return toVal
     return var
 
- def smFace(selItem, obj) :
+def smFace(selItem, obj) :
   # find face
   if type(selItem) == Part.Edge :
     Facelist = obj.ancestorsOfType(selItem, Part.Face)
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -397,7 +397,7 @@ class SMViewProviderFlat:
 
 
 class SMBendWallTaskPanel:
-    '''A TaskPanel for the facebinder'''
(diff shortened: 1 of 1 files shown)


shaise:
I tested your changes and I get an error when loading the workbench:
('unindent does not match any outer indentation level', ('C:\\Users\\shaise\\Downloads\\FreeCAD_0.17\\Mod\\FreeCAD_SheetMetal\\SheetMetalCmd.py', 109, 28, ' def smFace(selItem, obj) :\n'))
  File "<string>", line 44, in Initialize

It looks like there are identation errors. Do you use tabs? if so, use spaces instead. Can you check?

shaise:
Can you explain what this "selection based Edge" does?
When I select an edge and press Bend, I get an error:
  File "C:\Users\shaise\Downloads\FreeCAD_0.17\Mod\FreeCAD_SheetMetal\SheetMetalCmd.py", line 272, in execute
    extend1 = fp.extend1.Value, extend2 = fp.extend2.Value, kfactor = fp.kfactor, selFaceNames = fp.baseObject[1], MainObject = fp.baseObject[0])
  File "C:\Users\shaise\Downloads\FreeCAD_0.17\Mod\FreeCAD_SheetMetal\SheetMetalCmd.py", line 130, in smBend
    selFace = smFace(selItem, MainObject.Shape)
  File "C:\Users\shaise\Downloads\FreeCAD_0.17\Mod\FreeCAD_SheetMetal\SheetMetalCmd.py", line 112, in smFace
    Facelist = obj.ancestorsOfType(selItem, Part.Face)
<type 'exceptions.ReferenceError'>: This object is immutable, you can not set any attribute or call a non const method

PR #57, opened 2019-02-23
[WIP] Bend at sketch line

reox:
This was some code proposed in the sheetmetal thread (https://forum.freecadweb.org/viewtopic.php?f=22&t=25281&start=70#p245071)
but was never added into the repo.

The user Wega from FC forum did some archeology and found the code, which I added into git.

There are a few issues in the code, which should probably be resolved before merging.
Also the thing does not work in certain cases - so it probably needs some testing too.

### DIFF
--- a/SheetMetalFoldWall.py
+++ b/SheetMetalFoldWall.py
@@ -140,10 +140,12 @@ class SMFoldWall:
         obj.addProperty("App::PropertyLength","radius","Parameters","Bend Radius").radius = 1.0
         obj.addProperty("App::PropertyBool","invert","Parameters","Invert bend direction").invert = False
         obj.addProperty("App::PropertyAngle","angle","Parameters","Bend angle").angle = 90.0
+        # FIXME: I think this one does not work...
         obj.addProperty("App::PropertyBool","unfold","Parameters","Invert bend direction").unfold = False
         obj.addProperty("App::PropertyFloatConstraint","kfactor","Parameters","Gap from left side").kfactor = (0.45,0.0,1.0,0.01)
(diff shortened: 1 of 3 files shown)


SheetMetalFoldWall.py
```diff
@@ -0,0 +1,251 @@
...
+                wallSolid.rotate(revAxisP, revAxisV, bendA)
+                resultSolid = resultSolid.fuse(wallSolid)
+                bendSolid = cutFace.revolve(revAxisP, revAxisV, bendA)
+                #bendSolid = makeBend(BFace, bendR, bendA)
+                #Part.show(bendSolid)
+                resultSolid = resultSolid.fuse(bendSolid)
+        else :
+            if bendA > 0.0 :
+                # create bend
+                # FIXME undefined revFace!
```
reox, line 126: Not defined, looks like copy paste relict?

SheetMetalFoldWall.py
```diff
@@ -0,0 +1,251 @@
...
+
+class SMFoldWall:
+    def __init__(self, obj):
+        """Add Wall with radius bend"""
+        selobj = Gui.Selection.getSelection()
+
+        obj.addProperty("App::PropertyLength","radius","Parameters","Bend Radius").radius = 1.0
+        obj.addProperty("App::PropertyBool","invert","Parameters","Invert bend direction").invert = False
+        obj.addProperty("App::PropertyAngle","angle","Parameters","Bend angle").angle = 90.0
+        # FIXME: I think this one does not work...
```
reox, line 143: there are three things with description "Invert bend direction": unfold, invert and invertbend.
Not sure which of them is the correct one or what the others do...

reox:
As #58 was merged, I think I'll close this one ;)

PR #58, opened 2019-03-07
Fold wall

jaisekjames:
Fold sheet metal wall along a sketch.

Step to use  
1. select a face to bend 
2. then select sketch
3. Press icon
4. select inverbend to flip selection side

Feedback & grammatical corrections are welcome. 
It is based on fillet. so where it fail, this also fail.
unfold not implemented.


shaise:
see my comments on the forum.

jaisekjames:
updated based on comments in the forum

Issue #68, opened 2019-04-20
Definition of k-factor


ceremcem:
Internet searches as well as [Wikipedia page](https://en.wikipedia.org/wiki/Bending_(metalworking)#K-factor) shows that the definition of "k-factor" is as follows: 

> K-factor is a ratio of location of the neutral line to the material thickness as defined by t/T where t = location of the neutral line and T = material thickness. (...) and usually between 0.3 and 0.5. 

Based on this information, [latest change of k-factor in SheetMetal](https://github.com/shaise/FreeCAD_SheetMetal/commit/62c40b14fe6fea982b0b90c0490ea417b79cef51) is misleading, thus should be reverted.

Moreover, the formula [`newNode.k_Factor = (0.65 + 0.5*math.log10(innerRadius/self.__thickness))`](https://github.com/shaise/FreeCAD_SheetMetal/blob/bf2c0c1ecbda40ceedda646c1fdf63b39810c896/SheetMetalUnfolder.py#L828) has no explanation. It should be explained or removed. There is a long discussion about this formula in SW forum and seems that it's still unclear: https://forum.solidworks.com/thread/38306

shaise:
Hi Ceremcem,
Yes... This issue game me a headache for quite sometime... 
I will answer here both on this issue and your other issue #54. 
The unfold macro is quite old - it predates the sheetmetal workbench. It was created by ulrich1a, and I later added it into the workbench. As a separate macro, it does not rely on any of the sheetmetal object history or properties and has it own, single k factor usage. 
The K factor usage in this macro was indeed different then the one in the wikipedia (by a factor of 2) so I modified his code to conform to the wikipedia.
About a year ago, another programmer joined the workbench, jaisekjames. He added a feature for that lets you set an individual K factor to each fold. To use it you need to set the unfold property of the fold to true, and this fold will be straighten. 
Of course, to unfold the entire model, you will need to set the unfold property one by one for all folds. 
Then someone found out that doing that (setting the unfold property of all folds to true) results in an object different then the one generated using Ulrich's macro. Turns out jaisekjames used the same k-factor calculations as Ulrich's original.
After contacting Ulrich using the forum, I was informed there are 2 standards for K factor (european and american, dont remember which is which) So I decided to revert the K factor to Ulrich's original, both because it is his code, and because it is compatible to jaisekjames's code.
[...]

PR #69, opened 2019-04-21
Cca

ceremcem:
* Reverted the definition of K-factor 
* Tidy up the console output 
* Unfold operations uses the K-factor defined in Bend properties. 

Related to #68


shaise:
why remove the auto k-factor formula?

ceremcem:
@shaise Because this formula is used to determine the K-factor and K-factor is requested from user with an input in the Bend properties. That's why this introduces a UX conflict: Should K-factor be determined by the user, or calculated by SheetMetal WB? It's obvious that the determination of K-factor is tightly related with material properties, so it can't be calculated with such a formula. Besides, this formula needs a citation. The closest formula I can find on the net is the following: 

image

At the best case, this or any other accurate formula should be supplied as a separate tool and actual K-factor should be requested from the user.

shaise:
K-factor can not be taken from the model for 2 reasons:
1. there can be different factors for different bends (usually common when using different bend angles)
2. model can be imported / generated without any folding properties.
Regarding the formula, I agree that maybe it should be some kind of an option, or pop up calculator with selection of common materials.
Regarding ANSI / DIN, perhaps the best option is just to add a global Sheet Metal Workbench preference.

ceremcem:
> Regarding ANSI / DIN, perhaps the best option is just to add a global Sheet Metal Workbench preference.

I agree. This setting should be per project and if it's not explicitly set (then it is a project that uses an older version), user would be clearly informed about the difference and will be required to make the decision. If this preference is set globally, then only the newly created projects will use the global preference.

> there can be different factors for different bends (usually common when using different bend angles)

That is obvious. However, I'm not sure if I could understand this statement as I didn't break anything related to this usage. 

> model can be imported / generated without any folding properties

I didn't understand this one too.

ceremcem:
Oh, I get it now, you mean "K-factor can not be extracted/obtained from the model". 

> there can be different factors for different bends (usually common when using different bend angles)

Currently if user defines the same K-factor for each bend, this value will be used. Otherwise, a proper [error message is displayed](https://github.com/ceremcem/FreeCAD_SheetMetal/blob/ced509a0835aa6a4f3b172ed69963fb0de09ba7e/SheetMetalUnfolder.py#L2181) and operation is aborted. User is still able to follow the "unfold bend by bend" approach though. Currently I don't know how to handle multiple k-factor values but I'm planning to add this support ASAP. Anyway, this is not meant to be a breaking change. 

> model can be imported / generated without any folding properties

Then the user will have to use the "Manual K-factor" input while unfolding.

ceremcem:
@shaise Please let me test more cases as the parameter extraction is not working if the Body is somewhere deep inside several boolean operations.

shaise:
Personally I do not like the fold-by-fold unfold system. It is extremely limited, and can easily brake the model when other features are added. (Which is very common in sheet metal. For example a simple added hole through in the model, disappear after fold-by-fold unfolding:
Before:
bendunfold1
After:
bendunfold2

Using Ulrich's code to unfold will of course work without an issue.
I know that Ulrich was planning to make is code able to handle per-bend K factor, but I don't know the status.
The best way as I see it is to be able to click on bends using the gui, and set the K factor for each bend. This way it will support and sheetmetal object, not only ones generated by the workbench.

ceremcem:
I'm closing this PR regarding to #70

Issue #54, opened 2019-01-23
Unfold option Issue


jaisekjames:
See Issue

https://forum.freecadweb.org/viewtopic.php?f=22&t=25281&start=150#p279502
please change from
unfoldLength = ( bendR + kfactor * thk / 2.0 ) * bendA * math.pi / 180
to 
unfoldLength = ( bendR + kfactor * thk / 2.0 ) * bendA * math.pi / 180.0
Add decimal place to 180.

shaise:
Thanks! I will take a look

shaise:
Hi,

The 180.0 is not the problem. The problem is difference between European and American standard of k_facror (one of them is 1/2 the other). I have reverted it to the European standard, which is compatible with Ulrich's original unfolder script. Now both unfolding from your system, and Ulrich's system match.

jaisekjames:
Thank you

f3nix:
Thanks! :)

I have tested the changes and here are my observations.

Cases 1 and 3 work as expected now. :+1: 
I have a question about case 2. Should it use the kfactor set in the kfactor parameter (property).

![image](https://user-images.githubusercontent.com/72836/51693628-a19c3000-1fff-11e9-8257-86d4aa312217.png)

It seems it uses 0.65 in this instance.

`up Face7 k-factor: 0.65`

Cheers,
Mateusz

shaise:
The k factor in the properties are total different system and works only on sheetmetal prefolded using the sheetmetal WB. If you make a sheetmetal model using another system, you can only unfold it using the unfold button. The unfold button uses its own K-factor, and its the same factor for all folds. The properties system can set a different k factor to each fold.

f3nix:
Ok. Here is what I'm doing.

1. Create body and sketch.
2. Draw a rectangle (82mmx90mm).
3. Pad it (4mm).
4. Create a bend.
5. Change default properties:
   * length: 82mm
   * radius: 4mm
   * kfactor: 0,86
   * reliefd: 0mm
   * reliefw: 0mm
6. Select face and click "Flatten folded sheet metal object".

The unfolded length is **182,33mm** (which is different than other two unfold cases).

Report view says:
`up Face7 k-factor: 0.65`

It does _not use_ the **0,86 kfactor** that I have set in the bend property.

Is it the intended behaviour (workflow) or am I doing something wrong?

[...]

PR #70, opened 2019-04-22
K factor from lookup

### LINKED ISSUE #40: How to make unfold sketch updated on base geometry changes
This is the current unfold output: 

image

When I changed the geometry, for example as in the following way: 

image

I expect the unfold sketch to be updated accordingly. How can I achieve this, either automatically (this is preferred) or by manually running some code?

### LINKED ISSUE #68: Definition of k-factor

ceremcem:
# Description 

This PR adds the "Material Definition Sheet" concept to SheetMetal. 

K-factor is now obtained:
* either from manual K-factor input (mandatory if you don't have a material definition sheet)
* or from a "Material Definition Sheet"

# Manual K-factor

Users are now required to click and enable the "Manual K-factor" checkbox each time they want to unfold. This approach makes us quite confident that the user is seriously warned about the "`kfactor`" property from any `Bend` operation is not being used, so user won't encounter an unexpected result because of a default K-factor:

k-factor-manual

# Using a Material Definition Sheet 

k-factor-lookup2

K-factor property of a Bend is **completely ignored** since K-factor has nothing to do with the geometric data. If it is required to use different K-factor values per bend, follow the below steps: 

1. Determine your material name (eg. `foo`)
2. Add `_material_foo` postfix to your shape 
3. Create a spreadsheet with the name of `material_foo`
4. Create a table layout in `material_foo`, like the following (see [this table](https://user-images.githubusercontent.com/6639874/56498031-b017bc00-6508-11e9-8b14-6076513d8488.png)): 

    | Radius / Thickness | K-factor | Options | | 
    | ---| ---| --- | --- |
    | 1 | 0.38 | K-factor standard | ANSI |
    | 3 | 0.43 | | |
    | 99 | 0.5 | | |

5. Unfold as usual

# Is Breaking?

No.

# Changelog 

* K-factor standard is required to be explicitly set. (closes #68, related to #54 )
* Terminology is added to the README (closes #68)
* Material definition sheet is required unless manual K-factor is being used (new feature). 
* Added "Update Unfold" macro which resolves #40
* Fixed some typo
* Cleanup console output


shaise:
I like this approach much better!

However, I don't like the need to append the spreadsheet name to the bend.
This means that each time I add new feature to the model, I need to rename it.
I think it is better to create a spreadsheet with a known prefix, say "fold_material_xxx",
When you press unfold, it will take this spreadsheet automatically, and if there is more then
one, a drop down list will let me select.
If there is no spreadsheet, the manual K factor will be automatically selected. (in fact it can be
one of the options in the dropdown: Fold_material_xxx, fold_material_yyy, ...,  Manual.)

shai

ceremcem:
> I like this approach much better!

Thanks. 

> This means that each time I add new feature to the model, I need to rename it.

It will be frustrating to rename the model each time, yes. However, it becomes unnecessary as your model *gets more complex* because, 

1. Let's assume that you created a Body and created some bends. You need to rename it to `Body_material_foo` for the first time, as normal. 
2. If you add more bends, you *don't need to rename* the model. 
3. If you apply a boolean operation, new name becomes, eg. Fusion; so you **need to rename** it. 
4. If you want to fuse more solids, just drop into the `Fusion` container, you *don't need to rename*.
5. If you create a Cut, you **need to rename**. 
6. If you create more cuts, just [replace the Cut tool with a Fusion of cut parts](https://forum.freecadweb.org/viewtopic.php?f=3&t=35689&p=301744#p301744), so you *don't need to rename*. 
7. After this point, you *won't need to rename anything anymore* as you can add any solid you want to cut with into `Cut.Tool.Fusion` and any solid you want to fuse with into `Cut.Base.Fusion`. 

Moreover, you would only rename the model one or two times because you would probably want to unfold accurately after your model is completed, so you would rename the resulting model only after your work is completely done. Before your work is completed, you can always test unfold by providing the manual K-factor.

It's quite possible to provide a dropdown to be used in manual operations. However, a problem I can think of is that providing a manual K-factor as well as a material definition sheet manually means the user is providing a crucial information manually, which means the project file depends on a serious information that only the user memorizes. Since a project won't be considered as "Completed" unless such a critical information can be stored within the project file, there *MUST* be a way to store that information. If we could store this information, then it becomes obsolete to provide such a manual selection interface.

shaise:
Hi,

Sorry, but I really think that renaming an object as a way to add a
property into it is kind of a hack. It also very not intuitive to a user. A
user is doing good with selecting things and clicking buttons. Also, say
there is another workbench using the same technique, what then? rename the
object to Body_meterial_foo_otherproperty_bar ?

shai

ceremcem:
> I really think that renaming an object as a way to add a
property into it is kind of a hack.

You are absolutely right. It's not a correct way to go considering the FreeCAD UX way, it's just a hack. 

It turns out that this kind of necessity might have lead us a new feature of FreeCAD: Propagated Properties (@realthunder would you comment on this?). A "Propagated Property" will be like any other property of the root object. If the object is encapsulated with some other operator (like Bend, Fuse, Cut, etc.) the property will be removed from its original location and added to the root object. Exploding a root operation (such as deleting a Cut) will result in moving the Propagated Properties (back again) to the newly created root object which was the `Base` object before explosion.

ceremcem:
Latest commit adds the ANSI/DIN selection radio box, thus removes the "breaking" status of the PR: 

image

It's also required to declare `K-factor standard` explicitly in the `Options` within Material Definition Sheet: 

image

As the "Label renaming hack" is harmless and is only used to add a new feature, this PR is ready to be reviewed.

shaise:
Having a selection of DIN/ANSI is a great Idea to settle this issue. 
I will review now the changes.

shaise:
Hi ceremcem,

I took a look at it, and here is what I think:
First of all There is no problem to merge this change. The only lost feature is the automatically calculated K factor which I agree that in most cases there is no use for it.
The big concern however, is that your complicated changes which are great will be lost without use. When I try to unfold I get this message telling me to set a material definition sheet. That is great, but there is no way a user might guess how to do it. He will probably look hopelessly in the menu's to see where this "Material Sheet" is hiding.
There must be some kind of an online help connected to this message to explain how to do it. Or perhaps a button on the toolbar that generates a template sheet.

ceremcem:
Do you think above instructions in the very first post of this PR is sufficient so I could add it to the README and point this document in the error message, or should it be more clear?

shaise:
I think for a start its definitely good enough!

ceremcem:
Added. Please see 7f14194

ceremcem:
FYI: "Propagated Properties" issue has been risen in the FreeCAD forum: https://forum.freecadweb.org/viewtopic.php?f=3&t=35889

Issue #78, opened 2019-04-27
Add Material Definition Sheet selection combobox


ceremcem:
Related to #75 #70

ceremcem:
Resolved by #79

Issue #72, opened 2019-04-24
How to update Unfold Options menu via UnfoldOptions.ui?


ceremcem:
It turns out the `UnfoldOptions.ui` is not interpreted directly. AFAIU `pyuic4` is used to generate Python codes located in `SheetMetalUnfolder.py/SMUnfoldTaskPanel`. However, there are lots of code seems to be diverged from the output of `UnfoldOptions.ui`. 

What is the correct way to edit the UI?

shaise:
You are correct, this is my mess. I don't remember why I could not use the .ui directly, but this is the result. when I want to change it, I use pyuic4, then copy/paste over what the current lines are: 2162-2251

ceremcem:
I got it, it's OK. 

Regarding to using the `.ui` directly; I was using PySide a few years ago and [here](https://github.com/ceremcem/ggj-enzyme-game-backend/blob/da1b29338504fbfbbc6a5f25fe8d5eb01974b55c/display.py#L11) is the relevant part from one of my sample projects. The `Qt` wrapper is located [here](https://github.com/aktos-io/aktos-dcs-lib/blob/master/Qt/__init__.py). We may attempt to refactor the code to use `.ui` directly all in good time.

PR #79, opened 2019-04-27
Add mds combobox

ceremcem:
Related to #78 

image

Now users are able to: 
* List available Material Definition Sheets by the combobox. 
* Set any available MDS as well as remove the current MDS assignment. 

# Changelog 

* Added Material Definition Sheet selection combobox to the Unfold Task Panel. 
* Improved error outputs 

# Is breaking?

Nope.

### DIFF
--- a/Macros/SheetMetalUnfoldUpdater.FCMacro
+++ b/Macros/SheetMetalUnfoldUpdater.FCMacro
@@ -62,10 +62,13 @@ for o in doc.Objects:
 							# This view uses our unfolded output, update this view at the end 
 							echo("* Found related TechDraw ProjGroup: %s\n" % view.Label)
 							related_views.append(view)
-							
+			
+			# mark current output as "to be deleted"
+			tmp_postfix = "_TMP"
+			old = None		
 			try:
-				old = doc.getObjectsByLabel(output_name)[0].Name
-				doc.removeObject(old)
+				old = doc.getObjectsByLabel(output_name)[0]
+				old.Label += tmp_postfix
 				doc.recompute()
 			except:
 				pass
@@ -82,22 +85,35 @@ for o in doc.Objects:
 			Gui.Selection.clearSelection()
 			Gui.Selection.addSelection(doc.Name, unfold_face_id[0], unfold_face_id[1])
 			Gui.runCommand('SMUnfoldUnattended')
-			
-			# Remove unnecessary unfolded solid
-			doc.removeObject('Unfold')
-			
-			# Properly rename the output
-			sketch = doc.getObjectsByLabel('Unfold_Sketch')[0]
-			sketch.Label = output_name
-			sketch.Visibility = False
-			
-			# Update the source of related views
-			if len(related_views) > 0:
-				for view in related_views:
-					echo("* Updating %s source as %s\n" % (view.Label, sketch.Label))
-					view.Source = [sketch]
-			else:
-				warn('No TechDraw view is updated.\n')
+		
+			try:
+				# Remove unnecessary unfolded solid
+				doc.removeObject('Unfold')
+				
+				# Properly rename the output
+				sketch = doc.getObjectsByLabel('Unfold_Sketch')[0]
+				sketch.Label = output_name
+				sketch.Visibility = False
+				
+				# Update the source of related views
+				if len(related_views) > 0:
+					for view in related_views:
+						echo("* Updating %s source as %s\n" % (view.Label, sketch.Label))
+						view.Source = [sketch]
+				else:
+					warn('No TechDraw view is updated.\n')
+
+				# remove the temporary object
+				if old is not None:
+					doc.removeObject(old.Name)
+					doc.recompute()
+			except:
+				warn('Something went wrong.\n')
+				# restore the previous sketch
+				if old is not None:
+					echo("* Restoring the previous sketch.\n")
+					old.Label = old.Label[:-(len(tmp_postfix))]
+					doc.recompute()
 
 doc.recompute()
 Gui.Selection.clearSelection()
--- a/Macros/SheetMetalUnfoldUpdater.FCMacro
+++ b/Macros/SheetMetalUnfoldUpdater.FCMacro
@@ -54,8 +54,6 @@ for o in doc.Objects:
 						try: 
 							_src = view.Source[0]
 						except Exception as e:
-							warn("!! Exception of %s is: %s" % (view.Label, e.args[0]))
-							print(view.Source)
 							continue
 						#print "source of ", view.Label, ' in ', x.Label, " is: ", _src.Label
 						if _src.Label == output_name:
--- a/README.md
+++ b/README.md
@@ -44,11 +44,11 @@ To use the Material Definition Sheet, follow the steps:
 3. Create a spreadsheet with the name of `material_foo`
 4. Create a table layout in `material_foo`, like the following (see [this table](https://user-images.githubusercontent.com/6639874/56498031-b017bc00-6508-11e9-8b14-6076513d8488.png)): 
 
-    | Radius / Thickness | K-factor | Options | | 
-    | ---| ---| --- | --- |
-    | 1 | 0.38 | K-factor standard | ANSI |
-    | 3 | 0.43 | | |
-    | 99 | 0.5 | | |
+    | Radius / Thickness | K-factor | | Options | | 
+    | ---| ---| --- | --- | --- |
+    | 1 | 0.38 | | K-factor standard | ANSI |
+    | 3 | 0.43 | | | |
+    | 99 | 0.5 | | | |
     
     Notes: 
     
--- a/README.md
+++ b/README.md
@@ -35,14 +35,12 @@ You can use a Spreadsheet object to declare K-factor values inside the project f
 
 * Different K-factor values to be used for each bend in your model 
 * Sharing the same material definition for multiple objects 
-* Unattended unfold (required for ["Unfold Updater" macro](./Macros/SheetMetalUnfoldUpdater.FCMacro))
 
 To use the Material Definition Sheet, follow the steps:
 
 1. Determine your material name (eg. `foo`)
-2. Add `_material_foo` postfix to your shape's Label.
-3. Create a spreadsheet with the name of `material_foo`
-4. Create a table layout in `material_foo`, like the following (see [this table](https://user-images.githubusercontent.com/6639874/56498031-b017bc00-6508-11e9-8b14-6076513d8488.png)): 
+2. Create a spreadsheet with the name of `material_foo`
+3. Create a table layout in `material_foo`, like the following (see [this table](https://user-images.githubusercontent.com/6639874/56498031-b017bc00-6508-11e9-8b14-6076513d8488.png)):
 
     | Radius / Thickness | K-factor | | Options | | 
     | ---| ---| --- | --- | --- |
@@ -56,6 +54,7 @@ To use the Material Definition Sheet, follow the steps:
     2. Possible values for `K-factor standard` is `ANSI` or `DIN`. 
     3. `Radius / Thickness` means `Radius over Thickness`. Eg. if inner radius is `1.64mm` and material thickness is `2mm` then `Radius / Thickness == 1.64/2 = 0.82` so `0.38` will be used as the K-factor. See [lookup.py](https://github.com/ceremcem/FreeCAD_SheetMetal/blob/k-factor-from-lookup/lookup.py#L46-L68) for more examples.
 
+4. Use "Unfold Task Panel" to assign the material sheet.
 5. Unfold as usual.
 
 [Here](https://user-images.githubusercontent.com/6639874/56642679-a749f600-6680-11e9-944a-82e447d9dc4e.gif) is a screencast in action.
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -2153,7 +2153,6 @@ if pg.IsEmpty():
     pg.SetString("bendColor","#c00000")
     pg.SetString("genColor","#000080")
     pg.SetString("intColor","#ff5733")
-    pg.SetBool("persistentManualKfact", 0)
     pg.SetString("manualKFactor", "0.40")
 
 class QColorButton(QtGui.QPushButton):
@@ -2359,7 +2358,10 @@ class SMUnfoldTaskPanel:
         sizePolicy.setHeightForWidth(self.checkKfact.sizePolicy().hasHeightForWidth())
         self.checkKfact.setSizePolicy(sizePolicy)
         self.checkKfact.setObjectName(_fromUtf8("checkKfact"))
+        # NOTE: use `.clicked.connect` instead of `.stateChanged.connect` if you want to
+        # implement a "rollback" feature
         self.checkKfact.stateChanged.connect(self.checkKfactChange)
+        # END OF NOTE
         self.horizontalLayout.addWidget(self.checkKfact)
         spacerItem = QtGui.QSpacerItem(40, 20, QtGui.QSizePolicy.Expanding, QtGui.QSizePolicy.Minimum)
         self.horizontalLayout.addItem(spacerItem)
@@ -2407,11 +2409,6 @@ class SMUnfoldTaskPanel:
         self.verticalLayout.addLayout(self.horizontalLayout_2)
         spacerItem = QtGui.QSpacerItem(20, 40, QtGui.QSizePolicy.Minimum, QtGui.QSizePolicy.Expanding)
         self.verticalLayout.addItem(spacerItem)
-        self.checkPersistentManualKfact = QtGui.QCheckBox(SMUnfoldTaskPanel)
-        self.checkPersistentManualKfact.setObjectName(_fromUtf8("checkPersistentManualKfact"))
-        self.checkPersistentManualKfact.clicked.connect(self.checkPersistentManualKfactClicked)
-        self.verticalLayout.addWidget(self.checkPersistentManualKfact)
-
         self.verticalLayout_2.addLayout(self.verticalLayout)
 
         if genSketchChecked:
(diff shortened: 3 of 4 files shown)


shaise:
if I try to unfold and there is no sheet available, the wb crashes:
```
  File "D:\Shai\FreeCAD\FreeCAD_0.18\Mod\FreeCAD_SheetMetal-add-mds-combobox\SheetMetalUnfolder.py", line 2745, in Activated
    taskd = SMUnfoldTaskPanel()
  File "D:\Shai\FreeCAD\FreeCAD_0.18\Mod\FreeCAD_SheetMetal-add-mds-combobox\SheetMetalUnfolder.py", line 2267, in __init__
    self.availableMdsObjects = [o for o in doc.findObjects('Spreadsheet::Sheet') if material_sheet_regex.match(o.Label)]

'Spreadsheet::Sheet' is not a valid type
```

shaise:
i still get the above error. try to unfold the attached simple file.
[testfoldsheet.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/3123928/testfoldsheet.zip)

ceremcem:
This seems to be a FreeCAD bug because `'Spreadsheet::Sheet'` is a valid type and this code works correctly when there is at least one spreadsheet in the document. 

Anyway, I'm going to workaround it by encapsulating with a try/except block.

ceremcem:
It should work as intended now.

ceremcem:
Ready to be reviewed.

easyw:
@ceremcem 
I've noticed that the choice to use 'manual' or 'spreadsheet' is not saved/restored.

Moreover I'm on Linux Mint 19 and it seems the colors of the sketches are not correctly displayed ... it could be a Qt5 issue (that was pre-existant)...
qt5-issue

OS: Linux Mint 19.1
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.18.1.
Build type: Release
Python version: 3.6.7
Qt version: 5.9.5
Coin version: 4.0.0a
OCC version: 7.3.0

ceremcem:
> I've noticed that the choice to use 'manual' or 'spreadsheet' is not saved/restored.

"Use MDS" should be stored naturally because "saving" renames the object label and "restoring" reads the label and tries to match with available sheets. If it doesn't work in this way, let me know. 

I've also noticed that "Manual K-factor" selection is not persistent across unfold jobs. However, I assume this is by intention because providing a manual K-factor is a "transient action" and is not a part of the project. That's why I think "Manual K-factor" should behave as is and should be avoided except testing/debugging purposes. 

> it could be a Qt5 issue

Possibly. I'm also on Linux (Debian 9) and things work as expected with Qt4.

easyw:
@ceremcem, I would intend that if you select MDS, or Manual-K, this should be saved into the config.
Next time you open a FC doc, you should have one of this check box selected as last choice.

Are you on Deb9 FC 18.1 with Py2 and Qt4? The default installation should be Py3 & Qt5...

ceremcem:
@easyw 

I'm using an AppImage: 

```
OS: Debian GNU/Linux buster/sid
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.17.13541 (Git) AppImage
Build type: None
Branch: releases/FreeCAD-0-17
Hash: 9948ee4f1570df9216862a79705afb367b2c6ffb
Python version: 2.7.6
Qt version: 4.8.6
Coin version: 4.0.0a
OCC version: 7.3.0
Locale: English/UnitedStates (en_US)
```

"Libraries" tab shows that "Python 2.7.6" and "Qt 4.8.6" is being used. 

> Next time you open **a FC doc**

MDS can not be simply set to "selected" for "any" FC document. This is by definition. Material Definition Sheet assignment is just like `.Label`s, the assignment is very specific to the sheet metal object. So it can not be restored before first assignment. 

Regarding to the Manual K-factor, I'm against the persistent behavior. Checking a box is not an important effort and is invaluable in terms of human error prevention (I don't trust myself, I'm considering current behavior as a good error barrier for myself). 

### Deeper explanation

I'm trying to understand your point of view. If a user considers his/her FC project file as "completed", he/she **has to make** MDS assignments because MDS is a part of physical dimensions of his/her model. If MDS isn't declared for an object, project will be considered incompleted. If project is incompleted, it's in WIP state. If an action (Manual K-factor) performs another operator's (MDS) duty (correct, it does) but still doesn't suffice to make the project "completed", then this action HAS TO be considered a transient/debugging action. As no debugging actions should be persistent, providing Manual K-factor shouldn't be persistent either.

shaise:
selecting manual k-factor is not saved across fc runs on windows as well

easyw:
> I'm trying to understand your point of view. If a user considers his/her FC project file as "completed", he/she has to make MDS assignments because MDS is a part of physical dimensions of his/her model. If MDS isn't declared for an object, project will be considered incompleted.

@ceremcem Not really... consider the case in which you work always with a standard sheet metal thickness and each bend has the same radius, then a manual k-factor assigned is enough to consider the project complete.
That is most of my user case and I think it could cover many users. No need for creating a mds.

shaise:
> Not really... consider the case in which you work always with a sheet metal thickness andd each bend has the same radius, then a manual k-factor assigned is enough to consider the project complete.
That is most of my user case and I think it could cover many users. No need for creating a mds.

+1. We must leave the simple case valid. (even default IMHO)

ceremcem:
@shaise Please wait for merging this PR, I've found some issues with LinkStage3 handling.

ceremcem:
@easyw 
> consider the case in which you work always with a standard sheet metal thickness and each bend has the same radius,

First of all, I'm assuming we are talking for a production environment and very high precision is required (to explain my situation: I have to produce bent parts with 50 µm tolerance at most). (If you don't need such a precision, then you are correct, since a K-factor of 0.40 would cover nearly all cases)

Let's assume that you have created 100 projects with the setup you described. If you decide to change the material you are using in a near future, you may not be able to remember which material you were using for a project after a long duration since the switch. 

Anyway, if you still think that the persistent Manual K-factor selection is a necessity, then I'm sure that we can figure out something.

easyw:
@ceremcem 
> Let's assume that you have created 100 projects with the setup you described. If you decide to change the material you are using in a near future, you may not be able to remember which material you were using for a project after a long duration since the switch.

I don't know about you, but I have a written procedure for each production, I don't trust the habit, but the procedure I developed during the productions.

> Anyway, if you still think that the persistent Manual K-factor selection is a necessity, then I'm sure that we can figure out something.

Definitively I would suggest you to include the saving option as asked above.
Thanks.

easyw:
@ceremcem 
> I'm using an AppImage:
...
Version: 0.17.13541 (Git) AppImage

Please consider to switch to 0.18.1 new stable release, which has many bug fixes and improved features.
Moreover is using Py3 and Qt5. Py2 is on its ending of life.
[FreeCAD_0.18-16110-Linux-Conda_Py3Qt5_glibc2.12-x86_64.AppImage](https://github.com/FreeCAD/FreeCAD/releases/download/0.18.1/FreeCAD_0.18-16110-Linux-Conda_Py3Qt5_glibc2.12-x86_64.AppImage)

ceremcem:
@shaise I've fixed the problem related to LinkStage3 features. It's ready for your consideration. 

@easyw Thanks for the link.

shaise:
still not saving manual mode

ceremcem:
> still not saving manual mode

Because I haven't implemented it yet, I'm on it :) It's good news if there is no issue left other than that.

ceremcem:
Please use the following setting to enable "Use manual K-factor whenever possible" mode:

image

easyw:
@ceremcem 
sorry but for me this extra checkbox is adding confusion instead of simplifying the process/choice...
Why simply not mutual excluding Manual k-factor if MDS is selected (and inversely)?
(they should be an option more than a checkbox)

ceremcem:
@easyw 

1. The [update unfold](https://github.com/shaise/FreeCAD_SheetMetal/blob/7f14194cbe088879dde335b300be706d7b66e0cc/Macros/SheetMetalUnfoldUpdater.FCMacro) relies on this "You should choose whether using manual K-factor or MDS" error.
2. Think this error like "You can't start your car without applying a force to brake pedal first" in new cars. It's an extra step, but it worths it. 
3. Without this error, no one can guess that a Material Definition Sheet is another way of applying K-factor settings. With this error, it's pretty clear that user may provide K-factor manually or via a MDS: 

image

4. In your specific case, if you are always using same material, then you don't have to change any of the default settings. What would possibly make the confusion? Moreover, you will possibly use the unattended unfold button in your case, so you won't even see the Unfold Task Panel.

shaise:
I too think the extra check box is very confusing. I even prefer not to have the manual-k saved (as before) then have this extra check box.

easyw:
@ceremcem 
it seems to me you are trying to force people to use your approach, but SM is in use from very long time and nobody had issues before.
I like to have more options, but please keep things easy to users to keep previous behavior.
I really would suggest you to let the manual k-factor to be saved as well as the mds option, just letting the user decide, without forcing any direction.
Moreover, because MDS and MKF are mutually exclusive options, they should be 'radio btn' and not 'checkbox' controls.

ceremcem:
> I even prefer not to have the manual-k saved

Do you think the user should enter the desired K-factor for each unfold task if a different K-factor value is needed from the default one?

Regarding to the confusion of extra checkbox: Thinking again, yes, it might be confusing because duty of the checkbox is unrelated to an "unfolding task", it's directly related to UX preference. So I'm going to remove the chexbox.

easyw:
> Do you think the user should enter the desired K-factor for each unfold task if a different K-factor value is needed from the default one?

I would stay with a single MKF or with the MDS for complex job, I wouldn't mix the approaches.

shaise:
I still fully agree with @easyw.
The manual option should be saved across FC runs. This is the most common
and simple use, and compatible with previous versions.

ceremcem:
> but SM is in use from very long time and nobody had issues before.

See #54 , it's the issue where MDS idea had to be risen. Without MDS, **there was and there is no way** to store the K-factor with the project file. It's the only way.

>  compatible with previous versions
> please keep things easy to users to keep previous behavior

Do I misremember the previous behavior? Manual K-factor selection had never been **selected** by default. Moreover, it's used by default even it's not selected, which was a serious sneaky UX **bug**. Wait, you are both correct if you mean "the bug was only having the Manual K-factor selection checkbox unchecked, not using the Manual K-factor by default". 

> I like to have more options, but please keep things easy to users to keep previous behavior

This PR doesn't introduce any breaking change. This PR only reads the object label, converts it to the combobox and renames the object when you changed the material. That's all. 

Please switch to master branch and see the "previous" behavior. You'll be asked to select either MDS or Manual K-factor:

image

What you both are requesting is a new UX feature and is *literally breaking* the "previous" version's UX.

easyw:
the updated behavior is clear:
1) chose a MKF
or
2) chose a MDF

otherwise it is not possible to unfold.
Then one or other option should be saved when chosen.

ceremcem:
> **Then** one or other option should be saved when chosen.

Yeah! That was what I was about to propose!

ceremcem:
@easyw @shaise Please test current behavior.

easyw:
@ceremcem 
I've tested your last repo update
https://github.com/ceremcem/FreeCAD_SheetMetal/tree/add-mds-combobox
and it seems the option is saved on user.cfg,
but it is not loaded when a new file is open/created.
BTW, I also noticed your https://github.com/ceremcem/sheetmetal_calculator
https://ceremcem.github.io/sheetmetal_calculator/
It would be nice to have it implemented in the WB 😄

ceremcem:
Hi @easyw 
>  it seems the option is saved on user.cfg

Which option is that? The "Manual K-factor" of the object or default Manual K-factor value?

> https://ceremcem.github.io/sheetmetal_calculator/

Glad you liked it. I prepared it for my manufacturers so we could easily obtain K-factor of a random material with a simple experiment and calculate precise "Punch Line Distance" on the fly. It would be very nice to have a "popup" option of `WebGui.openBrowser()` so we could follow DRY pattern. 

Anyway it's not that hard once we could use `.ui` file directly. #72

easyw:
I've tested the MKF option.

> It would be very nice to have a "popup" option of `WebGui.openBrowser()` so we could follow DRY pattern.

this is possible, but generally people don't like to be directed to web; an internal Dialog is more suggested.
Most WBs are using the ui in the same way it is used in SM... Qt design to design the ui and then import the code in python associating the functions.
Often it is faster in this way.

ceremcem:
MKF options seems to work as intended: 

man-k-factor

What is the exact problem you are experiencing?

ceremcem:
> generally people don't like to be directed to web; an internal Dialog is more suggested

Me neither. However, the page I've built can be run offline, so we don't need to redirect to web. Try right click, save and open via `WebGui.openBrowser('/home/ceremcem/a.html')` (Currently some dependencies are fetched from the web. However, it's easy for me to bundle them along with the page.)

easyw:
@ceremcem 
> What is the exact problem you are experiencing?

@easyw 
> but it is not loaded when a new file is open/created.

(and I don't have any MDS) 
at least with FC 0.18.1 on win/lin

easyw:
> However, the page I've built can be run offline, so we don't need to redirect to web. Try right click, save and open via `WebGui.openBrowser('/home/ceremcem/a.html')`

as far as I remember opening the browser through QT is not working on all platforms... 
Anyway I have an 'internal' link to a pdf inside my StepUp WB QtGui Help page

ceremcem:
> it seems the option is saved on user.cfg
> but it is not loaded when a new file is open/created.

My confusion was that the MKF isn't stored in user.cfg, it's a "per object" value. Anyway, it seems to work in 0.18 FC main:
k-factor-working-018

ceremcem:
[debug-k-factor.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/3128491/debug-k-factor.zip)

easyw:
> Anyway, it seems to work in 0.18 FC main:

not if you follow my previous suggestion
> but it is not loaded when a new file is open/created.

mkfnotsaved

ceremcem:
What a long PR negotiation is that :)) 

Well, above behavior is expected: 

> it's a "per object" value. 

# Reason for having K-factor per object 

A project may perfectly contain multiple sheet metal objects with different material. For example, my current product contains 

1. Aluminum AL5754H22 for case (gray, outer)
2. Plexiglass for inner socket background. (yellow, highlighted)

image

Those two materials will require different K-factors (or different MDS's, if preferred). That's why any K-factor assignment should be per object.  

# Risk of default K-factor assignment

If we accidentally forget to assign the actual K-factor or MDS accordingly, as there are no barriers left to warn us, we'll end up having a wrong prototype in the best case.

Think that like DRC in KiCAD. AFAIK, You can't leave a pin floating or DRC will throw an error. You have to mark that pin "NC" explicitly. 

# Current Behavior 

There is no difference between MDS or MKF handling at the moment. None of them is assigned to the object by default. User is required to select either one of them *for every single sheetmetal object*.

An engineering tool shouldn't decide such a crucial parameter by default. If you don't wear your safety glasses, it would be more comfortable to work and 99% of the time you won't get into trouble. However, I still vote for "safety first" approach. 

# A Real Story

Do you know what was the case that made me find #54 and [make this comment](https://github.com/shaise/FreeCAD_SheetMetal/issues/54#issuecomment-485136660)? Exactly that "default parameter is easiness" behavior. It cost my 3 days to overcome the issue (I had to go to the end of the queue) and another 3 days to take my time to fix the issue in SheetMetal. 

# Further Proposal 

The "Default Manual K-factor Assignment" is a dangerous feature and users should enable it at their own risk. If we can provide a "Settings" page (I don't know how to create it), I can move that "Use global K-factor by default" to the settings page. 

How does it sound?

easyw:
@ceremcem 
> MKF isn't stored in user.cfg, it's a "per object" value

Where do you save your "per object" settings? Inside the FC file? or whatever unless user.cfg?
And if you save it in user.cfg, you then are bloating the cfg with as many items as the SM FC files a user is designing.


> The "Default Manual K-factor Assignment" is a dangerous feature and users should enable it at their own risk. If we can provide a "Settings" page (I don't know how to create it), I can move that "Use global K-factor by default" to the settings page.

I've noticed recently you have introduced a button to click and Unfold without opening the K-Factor Dialog...
That is exactly what I would avoid because this is exactly a dangerous way to unfold, dropping the user attention away from settings. (this should be exactly what you are trying to avoid, but this is dropped only if the user is acting on 'your way').

Then I would suggest to:
1) remove the 'error prone' default settings unfold button that is hiding the unfolding dialog;
2) allow to users to stay simple and let to save the MKF globallyt and NOT "per object".

shaise:
> Then I would suggest to:
>     1. remove the 'error prone' default settings unfold button that is hiding the unfolding dialog;
>     2. allow to users to stay simple and let to save the MKF globallyt and NOT "per object".

+1

ceremcem:
> Where do you save your "per object" settings? 

In the `.Label`s? O_o

> That is exactly what I would avoid because this is exactly a dangerous way to unfold, dropping the user attention away from settings. (this should be exactly what you are trying to avoid, but this is dropped only if the user is acting on 'your way').

Currently it doesn't do anything harmful because you can't perform "unattended unfold" unless you set the K-factor explicitly. This button is safe at the moment regardless the usage of MKF or MDS.

ceremcem:
Here is the side effects of a global, default Manual K-factor: 

1. Currently MDS approach uses "Label hack" until [Propagated Properties](https://forum.freecadweb.org/viewtopic.php?f=3&t=35889&sid=b8774acab57edb03bd255a73dac3ad84) land in FC (if it will) Using global default MKF will drive the users errors when they add more features to the bent object. If default MKF is enabled, the MDS feature should be completely removed because it becomes dangerous.

2. #40 depends on "Unattended Unfold" API which takes care of unfolding regarding to the **K-factor per object**. Unless K-factor is assigned per object, this API becomes useless, so #40 should be reopened or labeled as "`wontfix`". 

3. The patch for your requirement is as follows: 

```patch
diff --git a/SheetMetalUnfolder.py b/SheetMetalUnfolder.py
index cbdaa73..6d80dac 100644
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -2445,6 +2445,9 @@ class SMUnfoldTaskPanel:
             self.kFactSpin.setProperty("value", manKFactor)
             self.checkKfact.setChecked(True)
             self.updateKfactorStandard(virtual_material["standard"])
+        elif self.material_sheet_name is None:
+            self.checkKfact.setChecked(True)
+         
 
     def get_root_obj(self):
         selobj = Gui.Selection.getSelection()[0]
```

However I don't want to take the responsibility. Add it at your own risk. 

Thanks for your consideration.

easyw:
@ceremcem 
we also miss the patch to remove the extra button you added to 'simplify' the way of working.

easyw:
@ceremcem 
Just to be sure, is the above patch getting the saved value for MKF or just assign it to checked if material_sheet_name is None?

ceremcem:
> we also miss the patch to remove the extra button

I forgot it, here it is:

```diff
diff --git a/InitGui.py b/InitGui.py
index 250e371..e2cb950 100644
--- a/InitGui.py
+++ b/InitGui.py
@@ -45,7 +45,7 @@ class SMWorkbench (Workbench):
         import SheetMetalUnfolder
         import SheetMetalBaseCmd
         import SheetMetalFoldCmd        
-        self.list = ["SMBase", "SMMakeWall", "SMExtrudeFace", "SMFoldWall", "SMUnfold", "SMUnfoldUnattended"] # A list of command names created in the line above
+        self.list = ["SMBase", "SMMakeWall", "SMExtrudeFace", "SMFoldWall", "SMUnfold"] # A list of command names created in the line above
         self.appendToolbar("My Commands",self.list) # creates a new toolbar with your commands
         # self.appendMenu("My New Menu",self.list) # creates a new menu
         # self.appendMenu(["An existing Menu","My submenu"],self.list) # appends a submenu to an existing menu
diff --git a/SheetMetalUnfolder.py b/SheetMetalUnfolder.py
index cbdaa73..6d80dac 100644
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -2445,6 +2445,9 @@ class SMUnfoldTaskPanel:
             self.kFactSpin.setProperty("value", manKFactor)
             self.checkKfact.setChecked(True)
             self.updateKfactorStandard(virtual_material["standard"])
+        elif self.material_sheet_name is None:
+            self.checkKfact.setChecked(True)
+          
 
     def get_root_obj(self):
         selobj = Gui.Selection.getSelection()[0]
```

> Just to be sure, is the above patch getting the saved value for MKF or just assign it to checked if 

It does assign a MKF (it's actually handled as a Virtual Material) if no assignment has been done before. For the rest of the unfold operations, it uses saved value per object.

ceremcem:
@shaise I'm reopening this PR as it adds only non-breaking features regarding to the previous release (V0.2.31). Please merge if no issues can be found in its own context. 

In case you think the previous PR (#70) introduced the breaking UX change we were talking all about, I did provide [the patch](https://github.com/shaise/FreeCAD_SheetMetal/pull/79#issuecomment-487889661) which you can freely apply after merging this PR.

easyw:
@shaise, are you evaluating to include this on SM code?
Thx 
Maurice

PR #83, opened 2019-05-14
adding a repairing tool for invalid tolerance values shapes

easyw:
- adding a repairing tool for invalid tolerance values shapes
- repairing sketch malformed

related thread
https://forum.freecadweb.org/viewtopic.php?f=22&t=25281&p=304555#p304555


easyw:
@shaise 
please have a look at this PR... I've added to sew a shape in case of invalid tolerance values after an offset operation and to 'clean' the generated outline sketch.

shaise:
I just read the forum thread. Nice work! thanks.

PR #102, opened 2019-10-21
README: show screencast instead of linking to it

luzpaz:
The way the text flows in this section of the README is hard to discern that there is a link to a screencast. I went ahead and just displayed it by default.


shaise:
Good Idea. thanks

PR #105, opened 2019-10-23
[WIP] Translation: Adding placeholders

luzpaz:
Not sure how to add translations here yet, so in the meantime putting comments where they need to be. Eventually these commits can be cherry-picked and merged. 

@shaise feel free to help if you can

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -49,7 +49,7 @@ class SMWorkbench (Workbench):
         import SheetMetalCmd # import here all the needed files that create your FreeCAD commands
         import SheetMetalUnfolder
         import SheetMetalBaseCmd
-        import SheetMetalFoldCmd        
+        import SheetMetalFoldCmd
         self.list = ["SMBase", "SMMakeWall", "SMExtrudeFace", "SMFoldWall", "SMUnfold"] # A list of command names created in the line above
         self.appendToolbar("My Commands",self.list) # creates a new toolbar with your commands
         # self.appendMenu("My New Menu",self.list) # creates a new menu
--- a/SheetMetalBaseCmd.py
+++ b/SheetMetalBaseCmd.py
@@ -50,6 +50,7 @@ def smIsSketchObject(obj):
 def smIsOperationLegal(body, selobj):
     #FreeCAD.Console.PrintLog(str(selobj) + " " + str(body) + " " + str(smBelongToBody(selobj, body)) + "\n")
     if smIsSketchObject(selobj) and not smBelongToBody(selobj, body):
+        # how to add QtCore.QT_TRANSLATE_NOOP() to this ?
         smWarnDialog("The selected geometry does not belong to the active Body.\nPlease make the container of this item active by\ndouble clicking on it.")
         return False
     return True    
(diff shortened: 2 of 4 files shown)


shaise:
I tried to read how translate internals work, but could not figure it out. Is it possible to move the translate inside SmWarnDialog, or must it always wrap each text string?

luzpaz:
@yorikvanhavre when you get a chance, care to weigh in?

luzpaz:
@vocx-fc do you mind shedding light on how we can do this ?

vocx-fc:
See this thread for reference, [Updated Localisation Page](https://forum.freecadweb.org/viewtopic.php?p=342336#p342336).

> I'm not entire sure what the differences are, but according to the Qt documentation https://doc.qt.io/qt-5/i18n-source-translation.html, QT_TRANSLATE_NOOP is used outside functions (variable, dictionary definitions, for example), and translate is used within functions (print statements, mostly).

I haven't investigated the differences deeply but for practical matters I would do it like this.

First define the functions
```python
if App.GuiUp:
    from PySide.QtCore import QT_TRANSLATE_NOOP
    from DraftGui import translate
else:
    def QT_TRANSLATE_NOOP(context, text):
        return text

    def translate(context, text):
        return text
```

Use `translate()` inside a print function, and `QT_TRANSLATE_NOOP()` only in dictionary definitions.

Examples,
```python
smWarnDialog("The selected geometry does not belong to the active Body.\n"
             "Please make the container of this item active by\n"
             "double clicking on it.")

smWarnDialog(translate("SheetMetal",
                       "The selected geometry does not belong to the active Body.\n"
                       "Please make the container of this item active by\n"
                       "double clicking on it."))
```

As you will be using the `context=SheetMetal` everywhere, I'd use a shorthand.
```python
def _tr(text):
    return translate("SheetMetal", text)
```
Then
```python
smWarnDialog(_tr("The selected geometry does not belong to the active Body.\n"
                 "Please make the container of this item active by\n"
                 "double clicking on it."))
```
For the properties
```python
obj.addProperty("App::PropertyLength","length","Parameters","Length of wall").length = 10.0

obj.addProperty("App::PropertyLength","length","Parameters",
          QT_TRANSLATE_NOOP("App::Property", "Length of wall")).length = 10.0
```
As I said, I haven't had time to test this exactly, but this is the way it's done in Draft, so by imitating this style, it should work. Once the translate functions are in place, if they don't work for some reason, we can always re-define them, but at least they are there.

See also [Translating_an_external_workbench](https://www.freecadweb.org/wiki/Translating_an_external_workbench).

shaise:
@vocx-fc, is it possible to use _tr inside smWarnDialog() function body
and then everything sent to smWarnDialog() will be translated?

vocx-fc:
Yes, although, I prefer to not do it. Why? Well, a philosophy of Python is that "explicit is better than implicit". If you see the `smWarnDialog(_tr("Something"))` then it is evident that this piece of text is meant to be translated. However, if you don't see the `_tr()` it is not clear if that piece of text is already translated or not.

So, for this case, I prefer to have `_tr()` explicitly there to show other programmers what is actually being translated. Otherwise, another programmer may add a second `translate()` function without realizing there is already one in the back.

shaise:
understand and agree. Further more if I want to show a parameter as well: 
smWarnDialog("Diameter must be more then " + str(diam))
this will totally fail translation.

luzpaz:
Closing in favor of #183  
Great job @jaisekjames

PR #109, opened 2020-02-03
py3.8 update: time.clock() -> time.process_time()

looooo:
for py2 time.process_time is defined in the FreeCADinit.py:
https://github.com/FreeCAD/FreeCAD/commit/6866397c14226a504232b027b73ce4085e73f01e

### DIFF
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -110,6 +110,11 @@ from lookup import get_val_from_range
 import Part
 import tempfile
 
+# Backward compatibility to Py2
+import sys
+if sys.version_info.major < 3:
+    import time
+    time.process_time = time.clock
 
 genSketchChecked = True
 genSketchColor = '#000080'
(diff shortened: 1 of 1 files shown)


shaise:
is it backward compatible?

wwmayer:
> is it backward compatible?

No. It only works for all FreeCAD versions based on Py3 and FreeCAD v0.19 with Py2. But it doesn't work for FreeCAD v0.18 or older based on Py2. In order to support them too you must add the same workaround as I did in the above commit in looooo's post.

shaise:
@looooo , can you add the workaround at the beginning of the file so it will be backward compatible?

looooo:
@shaise 
You are right. Added the workaround for py2.

shaise:
thanks! merged

shaise:
Tested on v0.17 w/Py2, works. Thanks!

PR #114, opened 2020-03-08
Fix #113 'Bug in "Create a sheetmetal wall from a sketch" wrong radius for open sketches'

### LINKED ISSUE #113: Bug in "Create a sheetmetal wall from a sketch" wrong radius for open sketches
When creating a sheetmetal wall from an sketch, the resulted radius is not the same as the one passed in parameters, because the middle wire filet radius computation is wrong, it's `radius + thk / 2` instead of `radius * 1.5`

See https://forum.freecadweb.org/viewtopic.php?f=3&t=43980

hl037:
The small radius should be `radius`, then the big radius should be `radius * 2`. Since the boundaries are obtained by offseting the middle sketch, then the right radius should the in between : `(radius + radius + thk) / 2 = radius + thk / 2`

### DIFF
--- a/SheetMetalBaseCmd.py
+++ b/SheetMetalBaseCmd.py
@@ -69,7 +69,7 @@ def smBase(thk = 2.0, length = 10.0, radius = 1.0, Side = "Inside", MainObject =
       else :
         wire = WireList
       #Part.show(wire)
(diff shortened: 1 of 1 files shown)


hl037:
see https://forum.freecadweb.org/viewtopic.php?f=3&t=43980

shaise:
Thank you! This is more consistent.

PR #117, opened 2020-03-17
Support geometry element mapping

realthunder:
It enables SheetMetal objects to use the new topological naming in my [fork](https://github.com/realthunder/FreeCAD), so that it can auto change the geometry reference when the model changes.


shaise:
I dont know if it a good idea to make the changes before your fork gets into main. Are there any plans for that?

realthunder:
I plan to upstream the topological naming in next release, 0.20. Besides, the changes in this PR should be backward compatible. It checks for the new method `mapShapes` provided by the shape, and does nothing if it is not found.

shaise:
Great, then. I will merge it. 
Topological naming can be a game changer for FeeCAD. Thanks!

PR #118, opened 2020-03-26
Fix visibility in case of external editing

realthunder:
In case the user is editing an external linked object, the active document may not be the owner document of the editing object.


shaise:
Sounds good. Thanks!

PR #123, opened 2020-04-25
Update SheetMetalFoldCmd.py

jaisekjames:
Based on comment on forum
https://forum.freecadweb.org/viewtopic.php?f=3&t=45519#p390723


shaise:
Thanks for the fix!

PR #127, opened 2020-05-19
Add a new feature Bend to solid

jaisekjames:
A simple tool to add bend to edge solid.

Example if you have L shape solid, you can add a bend to corner by selecting corner edge.

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -49,8 +49,9 @@ class SMWorkbench (Workbench):
         import SheetMetalCmd # import here all the needed files that create your FreeCAD commands
         import SheetMetalUnfolder
         import SheetMetalBaseCmd
-        import SheetMetalFoldCmd        
-        self.list = ["SMBase", "SMMakeWall", "SMExtrudeFace", "SMFoldWall", "SMUnfold"] # A list of command names created in the line above
+        import SheetMetalFoldCmd
+        import SheetMetalBend
+        self.list = ["SMBase", "SMMakeWall", "SMExtrudeFace", "SMFoldWall", "SMUnfold", "SMMakeBend"] # A list of command names created in the line above
         self.appendToolbar("Sheet Metal",self.list) # creates a new toolbar with your commands
         # self.appendMenu("My New Menu",self.list) # creates a new menu
         # self.appendMenu(["An existing Menu","My submenu"],self.list) # appends a submenu to an existing menu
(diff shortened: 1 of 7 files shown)


jaisekjames:
pls test it before merging.

shaise:
Hi @jaisekjames ,

This is a very nice feature!
I see 2 issues now:
1. it can sometimes leave leftovers in the corner (see image)
2. In the model view on the left, If the part is inside a body, the bends should be in the same indentation one under the other (last bend at the bottom) and not like in part workbench which is a diagonal tree.
bend_solid

attached freecad file.
[test_corner_to_bend.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/4649990/test_corner_to_bend.zip)

jaisekjames:
Thank you. Both fixed. pls test it.

shaise:
great! I'm on vacation now, will test it this weekend.

jaisekjames:
Added two more command. Corner Relief, Junction.
See this video. 
https://www.youtube.com/watch?v=F9Rg_6PkJZU&list=PLONdQc9bizscSCvQuoFVFVClVSTLve0wT&index=6

shaise:
@jaisekjames ,
exciting new features!
However I still get some errors. the "create bends on solids" still does no work properly, and though the other commands seem to work, they give many errors.
I am using ver 0.19
test_newf1

test_newf2

[test_new_features.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/4671997/test_new_features.zip)

jaisekjames:
Fixed. messages are due to 
https://forum.freecadweb.org/viewtopic.php?f=22&t=12966&sid=aa48cebf84a0a29f5a8e55392dda4dba

currently we need to create relief, then junction[so that it will form open sheet] then final create bend.

shaise:
Hi @jaisekjames ,
Thanks for the fix. One last small thing: the junction icon on the model tree is same as the bend.
Also consider changing Bend name to other name because it is a different feature then standard Bend. (this way scripts that rely on feature name will work correctly) 
Thank you!

jaisekjames:
Thank for checking. both bug fixed.

shaise:
Merged! thank you!

shaise:
I have also changed: addObject("Part::FeaturePython","Bend") to addObject("Part::FeaturePython","SolidBend") so it will not be the same feature as "Bend".

PR #130, opened 2020-06-08
Engineering mode

ceremcem:
# Changelog 

* Introduced "Engineering Mode" (see https://github.com/shaise/FreeCAD_SheetMetal/pull/79#issuecomment-487774249 and [forum topic](https://forum.freecadweb.org/viewtopic.php?f=8&t=36424))
* Simplified Material Definition Sheet syntax


shaise:
hi @ceremcem,
This is a good idea. However adding a file as a flag is a bit of a patch. It should be under edit->preferences. Right now SheetMetal workbench have no preferences page. So it needs to be added. You can find an example at fasteners workbench. The first time you select this workbench, it will add a special page in edit->preferences. Let me know if you want me to add this page.
shai

ceremcem:
> However adding a file as a flag is a bit of a patch 

I agree, it was just a hack. 

> It should be under edit->preferences

I agree with this one too, however I wasn't sure about how to make such a thing. Thanks for the Fasteners WB tip, I'll take a look. Let me see what I can do.

ceremcem:
One situation: This "Engineering UX" is not meant to be a SheetMetal (or any WB) specific property, but it is a general user preference. Shouldn't it be included in FreeCAD source code, rather than SheetMetal WB preference page?

shaise:
it might, but its not for me to say.... :)

ceremcem:
@shaise Could you please check it out? Note that I didn't touch the README/Release notes part intentionally.

shaise:
Hi @ceremcem ,
Looks like it works correctly. Thanks for this feature.

PR #131, opened 2020-06-30
Some improvements

ceremcem:
* Fixed my "mass unfold updater" script, which now works on all documents at once. 
* Reduced the verbose output in order to help catching any important messages during the unfold operation. (This might be controlled via "Debug mode?" checkbox within the "Preferences" section, but it seems it's unlikely to be needed in the near future)


shaise:
merged. thanks!

PR #132, opened 2020-06-30
Bug fix.

jaisekjames:
Revised all code. Now bend will work on all bend angles
Need testing.

### DIFF
--- a/SheetMetalFoldCmd.py
+++ b/SheetMetalFoldCmd.py
@@ -219,9 +219,9 @@ def smFold(bendR = 1.0, bendA = 90.0, kfactor = 0.5, invertbend = False, flipped
       #print([neturalRadius, neturalLength, unfoldLength, offsetdistance, scalefactor])
 
       #To get facedir
-      tool_faces = tool.extrude(normal * -thk)
-      #Part.show(tool_faces, "tool_faces")
-      cutSolid = BOPTools.SplitAPI.slice(FoldShape, tool_faces.Faces, "Standard", 0.0)
+      toolFaces = tool.extrude(normal * -thk)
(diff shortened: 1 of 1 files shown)


shaise:
this a very big code change, what was the bug? Does it affect loading of old files?

ceremcem:
Plus, "Need testing" part made me anxious.

jaisekjames:
Earlier version was based on fillet to create bend. Now it is based exactly same theory as unfolder macro. 
I am able generate bend solid wire. see link macro
[https://forum.freecadweb.org/viewtopic.php?f=22&t=47588#p412121](url)

But there is no way to create cylindrical faces from wire.  So I was used a work around by guessing top & bottom faces are same & cutcylinder face from it.

If we able to generate faces from wrapped wires, we can create any shapes with this code.

There is some function in OCC "BRepBuilderAPI_MakeFace cylface(cylinder_face, wire) - limit cylinder by wired edges". If some developer able look into it & if useful, exposed to python, we can generate any bend shape from our code.

jaisekjames:
You can check folding using attached file in 

[https://forum.freecadweb.org/viewtopic.php?f=22&t=47588#p412121](url)

shaise:
Did you try loading old files to check if the changes did not break anything?

jaisekjames:
I checked for files, but there is no old file saved.

It will not break old since selection Procedure [face & sketch] is same. But there will be variation in results & face number change issue can happen because cutting position changes. New result will be match with other cads results also.

One Issue is there, here we use Part.makeFilledFace.  Some rare cases, it will generate hill like shape instead of cylinder. It will appear as holes in bends. I will check any other function to do that better.

PR #133, opened 2020-07-03
keeping Sketch Name as Sketch Label

easyw:
keeping Sketch Name as Sketch Label, useful to have meaningful names when exporting the sketches to DXF

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -33,7 +33,7 @@ smWB_icons_path =  os.path.join( smWBpath, 'Resources', 'icons')
 global main_smWB_Icon
 main_smWB_Icon = os.path.join( smWB_icons_path , 'SMLogo.svg')
 
-SHEETMETALWB_VERSION = 'V0.2.42'
+SHEETMETALWB_VERSION = 'V0.2.43'
 
 class SMWorkbench (Workbench):
  
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -2974,7 +2974,7 @@ class SMUnfoldTaskPanel:
         docG = FreeCADGui.ActiveDocument
         p = Part.makeCompound(edges)
         try:
-            sk = Draft.makeSketch(p.Edges, autoconstraints = True)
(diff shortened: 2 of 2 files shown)


ceremcem:
Please wait for my input before merging.

easyw:
@ceremcem 
which is the issue in merging it?

ceremcem:
Hi, 

Sorry, I was precautious. I thought this change will rename the unfolded objects' names from `Unfold###` and `Unfold_Sketch###` to something "meaningful". Such a change would break the `SheetMetalUnfoldUpater.Macro` as it [relies on that names](https://github.com/shaise/FreeCAD_SheetMetal/blob/74ac80e76c9923db20e3d8f2bb7a090f821980b1/Macros/SheetMetalUnfoldUpdater.FCMacro#L141-L145). I tried your change and it still makes the SheetMetal produce new unfold objects as `Unfold` and `Unfold_Sketch`, which is fine. 

No problem on my side ATM. 

However, I didn't quite understand what exactly this naming goes. Is that name accessible from the DXF viewer?

easyw:
@ceremcem , probably you don't generate separate projection sketches from the object... 
if you export the Sketches to DXF, inside the DXF before my PR, the block would be named i.e. Sketch001, Sketch002 (useless names); 
after my PR the name of the blocks inside the DXF would be the same as the label generated by the script... 
i.e. 'Unfold_Sketch_**bends**', 'Unfold_Sketch_**Internal**', 'Unfold_Sketch_**Outline**' ... "meaningful" & useful information to give to the bend manufacturer.
(please note that Name and Label are different in FC internal use, moreover your macro is based on Label which I didn't touched).

shaise:
@easyw , are you sure these are the only changes? I see 2 added variables to a function but nothing that uses them.

easyw:
@shaise 
I confirm that the PR is just fine... the var 'name' was already present, but simply was not used by 'Draft.makeSketch' to pass its value.
Now the sketch gets the same Label and Name; this is useful, as I said, when exporting the sketches to DXF, to give a meaningful name to the blocks inside the DXF file to the bender.

shaise:
Great! thanks!

PR #136, opened 2020-07-09
Re-added permanent manual K-factor setting

ceremcem:
@shaise Your commit (08c68dbf) had broken the "permanent manual K-factor setting for unattended unfold operations" feature, so I re-added it to take effect if user is in "Engineering UX mode".


shaise:
Done. thanks!

PR #142, opened 2020-09-11
Import os.path in InitGui.py Initialize function

### LINKED ISSUE #138: Not loading on 0.18
Hello,

I can't load this workbench on 0.18.4 anymore:
```
name 'os' is not defined
  File "<string>", line 66, in Initialize
```
OS: Windows 10
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.18.4 (GitTag)
Build type: Release
Branch: releases/FreeCAD-0-18
Hash: 980bf9060e28555fecd9e3462f68ca74007b70f8
Python version: 3.6.6
Qt version: 5.6.2
Coin version: 4.0.0a
OCC version: 7.3.0
Locale: Portuguese/Brazil (pt_BR)

With CAELinux 2018 (Xubuntu 16.04 LTS) the result is the same, with Freecad 0.18.4 Py3/QT5 and Py2/QT4.

ASzc:
When using the workbench on FreeCAD 0.18.4 Linux AppImage, this error
prevents it from loading:

```
name 'os' is not defined
  File "<string>", line 67, in Initialize
```

I have no idea why the global import added in commit 206bd12 doesn't
work, but I tried several different things, and this change fixed the
error for me.

This (re) fixes #138

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -58,6 +58,7 @@ class SMWorkbench (Workbench):
         import SheetMetalRelief
         import SheetMetalJunction
         import SheetMetalBend
+        import os.path
         self.list = ["SMBase", "SMMakeWall", "SMExtrudeFace", "SMFoldWall", "SMUnfold", "SMMakeRelief", "SMMakeJunction", "SMMakeBend"] # A list of command names created in the line above
         if engineering_mode_enabled(): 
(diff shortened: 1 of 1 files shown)


shaise:
Indeed this is strange why the global os is not recognized. I have no linux to test it, but since it is a harmless change, I'm merging it. Thanks!!

PR #144, opened 2020-11-14
Fold Bug Fix.

jaisekjames:
now work all bend angle

Thanks for Chris_G

https://forum.freecadweb.org/viewtopic.php?f=22&t=52101


shaise:
thanks for the fix

PR #145, opened 2020-11-17
2 Bend corner Relief feature

jaisekjames:
Initial version. Need testing. 

How to Test
* Create a sheet metal with base & walls.
* Click on corner vertexes
* Corner relief created. Adjust sizes or shape or use a sketch.

This is designed with only type of 2 bend wall
Do you have any idea of which all places require corner relief?

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -58,8 +58,9 @@ class SMWorkbench (Workbench):
         import SheetMetalRelief
         import SheetMetalJunction
         import SheetMetalBend
+        import SheetMetalCornerReliefCmd
         import os.path
-        self.list = ["SMBase", "SMMakeWall", "SMExtrudeFace", "SMFoldWall", "SMUnfold", "SMMakeRelief", "SMMakeJunction", "SMMakeBend"] # A list of command names created in the line above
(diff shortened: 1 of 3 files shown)


jaisekjames:
screenshot
corner relief

hungerburg:
This would be indeed a handy feature. Below a picture, where flattening fails, because the relief is missing (Freistich in German), which should be there for real reasons anyway. The area marked red will be just a square in the flat part.

I could not get that with currently available parameters _gap_ (neg. value) and _relief_ (pos. value), as they only ever cut out from the base part, not from the extension created with the bend.

The Extra from this pull request here in my installation (daily from ubuntu ppa) just corrupts the body part or is invalid from the start, with the simplest bend at all. (Could that be, because the part is a part design parametric sketch?) Maybe I need more detailed instructions on how to use corner-relief?

Perhaps though, a simpler way for corner relief might be, to make ordinary relief carry through to bend extensions?

corner-relief-freistich

jaisekjames:
need to revise this feature

hungerburg:
Corner relief would make a valuable addition to the the sheet metal workbench in FreeCAD. The picture on top looks quite nice. Likely must be an extra tool, in the mean time I have seen that in one other CAD handled specially too.

jaisekjames:
refer new pull request https://github.com/shaise/FreeCAD_SheetMetal/pull/147

PR #150, opened 2020-12-03
Extend improve

jaisekjames:
* Split wall extrude into new file.
* added extrude with sketch.
* extrude if intersect we can cut it with a offset.

extend0

extend1
extend2


shaise:
Nice!!!

PR #157, opened 2020-12-17
Extend - add normal to face

jaisekjames:
Refer
https://forum.freecadweb.org/viewtopic.php?f=3&t=11303&start=250#p454120
Now cutting done normal to face of sheet-metal.


luzpaz:
Would it be possible to add some comments to the code so other devs could understand what they're reading?

jaisekjames:
added 

https://github.com/shaise/FreeCAD_SheetMetal/pull/158

PR #164, opened 2020-12-23
Sheet metal Forming

jaisekjames:
Forming tool Updation

First select sheetmetal face
then continue selection by pressing. select face to align on forming tool & faces are required to remove.
then press forming icon.
Adjust sheetmetal thickness if required.

In Partdesign, We have use shapebinder.

Added multiple forming using sketch with circle as center for forming
forming
forming2
rename file without .zip
[forming test2.FCStd.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/5734755/forming.test2.FCStd.zip)

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -61,8 +61,10 @@ class SMWorkbench (Workbench):
         import SheetMetalBend
         import SketchOnSheetMetalCmd
         import SheetMetalCornerReliefCmd
+        import SheetMetalFormingCmd
         import os.path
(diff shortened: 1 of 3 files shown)


shaise:
Very Nice!

But this can not be unfold right? Is there a warning that the result can not be unfold?

jaisekjames:
There is a suppress feature is available.  If set True, it will bypass this feature. Unfolding can be done.

supress feature

shaise:
Sounds good!

PR #176, opened 2021-01-31
SMWall bugfix & improvement

jaisekjames:
* Split Code into small functions.
* Fixed gap value calculation issues & added mingap to property editor to adjust the gap. 
* Added extend values into auto-mitre & max extend value. Based on maximum extend value, automatically wall extended keeping min gap value.
* Fixed gap, extend values keep propagating into bends for sketch based wall & bend.
* Fixed  First angle is not calculated in sketch based wall & bend fixed.
* Bend radius values now considered in sketch based wall for wall height calculation [except first line].

pls see the issues on 

https://forum.freecadweb.org/viewtopic.php?f=3&t=11303&start=300#p469246
https://forum.freecadweb.org/viewtopic.php?f=3&t=11303&start=290#p468615

### DIFF
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -118,7 +118,7 @@ def smRestrict(var, fromVal, toVal):
     return var
 
 def smFace(selItem, obj) :
-  # find face if Edge Selected
+  # find face, if Edge Selected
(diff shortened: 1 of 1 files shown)


shaise:
Thanks!

PR #195, opened 2021-05-02
Context menu edit added

jaisekjames:
* Added Context menu edit in tree view for better discovery of editing


jaisekjames:
Added edit to context menu.

shaise:
great idea! thanks!

PR #196, opened 2021-05-04
bug fix - wall length in less than thickness

jaisekjames:
fix issue in

https://forum.freecadweb.org/viewtopic.php?f=3&t=11303&start=350#p501604


shaise:
thanks!

PR #197, opened 2021-05-06
Fixed bug where MainObj was in someplaces called obj

mangelozzi:
Look slike some functionality was added a few hours ago which referred to `MainOject` on line 342 in SheetMetalCmd.py, however it was being referred to in the function as `obj`, hence it would crash with `MainObject` not recognized. Simply rename of `obj` to `MainObject` to be consistent with the rest of the code.


shaise:
good catch, thanks!

PR #205, opened 2021-07-03
fix #204 SheetMetal AddBase allow PartDesign::SubShapeBinder as source

### LINKED ISSUE #204: `SheetMetal AddBase` allow `PartDesign::SubShapeBinder` as source
currently [SheetMetal AddBase](https://wiki.freecadweb.org/SheetMetal_AddBase) does not accept a 'PartDesign::SubShapeBinder' as source -
the icon is inactive.

if you manually afterwards change the `.BendSketch` property of the `BaseBend` (`PartDesign::FeaturePython`) Object it is just working. 
so it is only a GUI / pre-selection check that is to limiting.

s-light:
add
`PartDesign::ShapeBinder` 
`PartDesign::SubShapeBinder` 
as allowed selected source

### DIFF
--- a/SheetMetalBaseCmd.py
+++ b/SheetMetalBaseCmd.py
@@ -2,25 +2,25 @@
 ###################################################################################
 #
 #  SheetMetalBaseCmd.py
-#  
+#
 #  Copyright 2015 Shai Seger <shaise at gmail dot com>
-#  
+#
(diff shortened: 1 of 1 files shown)


s-light:
fix #204

shaise:
Thanks! looks good!

PR #209, opened 2021-08-05
fix #208

### LINKED ISSUE #208: <class 'AttributeError'>: 'NoneType' object has no attribute 'Visibility'
```
17:06:30  Traceback (most recent call last):
  File "/home/stefan/mydata/freecad/.FreeCAD/Mod/sheetmetal/SheetMetalBaseCmd.py", line 148, in execute
    Gui.ActiveDocument.getObject(fp.BendSketch.Name).Visibility = False
<class 'AttributeError'>: 'NoneType' object has no attribute 'Visibility'
```
this happens if i use a Binder that points to another document.
easie manual fix: just toggle the visibility of the object once.. 
i will write a PR soon.

s-light:
simply check if an obj exists before trying to access its propertie
this happens with partially loaded documents / linked things..

### DIFF
--- a/SheetMetalBaseCmd.py
+++ b/SheetMetalBaseCmd.py
@@ -145,7 +145,9 @@ class SMBaseBend:
                   midplane = fp.MidPlane, reverse = fp.Reverse, MainObject = fp.BendSketch)
 
     fp.Shape = s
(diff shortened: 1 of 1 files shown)


shaise:
Good catch! thanks!

PR #211, opened 2021-09-03
Bugfix SheetMetalExtendCmd

jaisekjames:
Fix issue discussed  on
https://forum.freecadweb.org/viewtopic.php?f=3&t=60818&start=430#p530117


shaise:
Thanks for the fix!

PR #212, opened 2021-10-04
Fix get_linked_objs_recursive()

realthunder:
Problem reported in https://forum.freecadweb.org/viewtopic.php?p=536616#p536616


shaise:
Thanks for the fix!!

PR #215, opened 2021-11-11
Extend feature not working when not passing through a wall.

jaisekjames:
bug fix for

https://forum.freecadweb.org/viewtopic.php?f=3&t=60818&start=500#p544183


shaise:
thanks!

PR #226, opened 2022-04-07
Remove version number appended to Sheet Metal wb name in menu dropdown

luzpaz:
This has been made obsolete by the new addon manager functionality that lists version info from package.xml in the About info dialog. Also it breaks uniformity esthetics with other workbench names + can be problematic with some themes like ModernUI, see example:

image

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -34,8 +34,6 @@ smWB_icons_path =  os.path.join( smWBpath, 'Resources', 'icons')
 global main_smWB_Icon
 main_smWB_Icon = os.path.join( smWB_icons_path , 'SMLogo.svg')
 
-SHEETMETALWB_VERSION = 'V0.2.49'
-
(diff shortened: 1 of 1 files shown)


shaise:
Good! I was about to suggest that myself

shaise:
But I think we need to bump the version to 50, so the addon manager will detect the change. no?

luzpaz:
Good point. Though Addon Manager detects the change nonetheless because it doesn't just look at the version number in the package.xml file. It uses git to ascertain if upstream repo has new commits. So it's your call you can bump or keep it as is

shaise:
Strange then, my addon manager did not detect the change

luzpaz:
Did chennes change that? Lemme check....

luzpaz:
Yea, it worked. Just refresh your addon manager cache

Issue #228, opened 2022-05-01
Can't bend something 2mm wide or narrower


DavidGriffith:
Sketch a rectangle measuring 20mm tall by 2mm wide. Then in the Sheet Metal workbench, select Make Base Wall. Now you have a sheet 1mm thick, 2mm wide and 20mm tall. Click a broad face and attach another sketch. Draw a bend line parallel to the 2mm sides. Now select a face and the bend line, then select Fold a Wall. This will fail with the following text printed in the report view and to standard error/output.

````
00:00:28  Traceback (most recent call last):
  File "/home/foobar/.local/share/FreeCAD/Mod/sheetmetal/./SheetMetalFoldCmd.py", line 259, in execute
    s = smFold(bendR = fp.radius.Value, bendA = fp.angle.Value, flipped = fp.invert, unfold = fp.unfold, kfactor = fp.kfactor, bendlinesketch = fp.BendLine,
  File "/home/foobar/.local/share/FreeCAD/Mod/sheetmetal/./SheetMetalFoldCmd.py", line 154, in smFold
    solid0 = solid1list[0]
<class 'IndexError'>: list index out of range
00:00:28  Recompute failed! Please check report view.
````

If I increase the width of this strip to 3mm or larger, then the bend is made without complaint.

Here's what my environment is like:

````
OS: Debian GNU/Linux 11 (bullseye) (MATE/lightdm-xsession)
Word size of FreeCAD: 64-bit
Version: 0.20.28774 (Git)
Build type: Unknown
Branch: master
Hash: d2253c7be4db12dbf9b89a53abfacff87692d216
Python 3.9.2, Qt 5.15.2, Coin 4.0.0, OCC 7.5.1
Locale: English/United States (en_US)
Installed mods: 
  * sheetmetal 0.2.49
  * kicadStepUpMod 10.11.9
  * fasteners 0.3.40
````

DavidGriffith:
Someone has confirmed this over at https://forum.freecadweb.org/viewtopic.php?p=592346#p592346

DavidGriffith:
Could I get some advice on what's going wrong here and how to fix it myself?

DavidGriffith:
#229 fixes the problem.  Thanks!  Closing now.

PR #229, opened 2022-05-06
Bugfix_fold

jaisekjames:
Smaller width sheetmetal is not folding due to thickness calculation wrong value.
Now estimated thickness based on selected face.

https://github.com/shaise/FreeCAD_SheetMetal/issues/228


DavidGriffith:
This fixes the problem for me.

shaise:
Thanks!

PR #231, opened 2022-05-27
Correct classname in package.xml

chennes:
You actually used `SMWorkbench` as your class in `InitGui.py`, so this updates package.xml to find it correctly.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -12,7 +12,7 @@
 
   <content>
     <workbench>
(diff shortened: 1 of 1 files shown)


shaise:
will it not confuse the add on manager?

chennes:
No - the classname entry must match your actual class name. The addon manager doesn't care, but the FreeCAD startup does (for locating your icon).

shaise:
thanks!

PR #234, opened 2022-07-09
moving from Drawing to TechDraw FC0.21 compatibility

easyw:
Since recently 'Drawing' has been deprecated (disabled) in FC main 0.21+,
this PR will switch the calls using Drawing to TechDraw, keeping SheetMetal fully functional both in FC stable and daily


shaise:
Thanks.

akshimassar:
Isn't it a bit early to merge? It's not working with 0.19 and realthunder branch either.

shaise:
@easyw, It seems that your fix is incompatible with ver 20 stable. Can you please check? see issue #235

PR #236, opened 2022-07-21
fix flange length in case bend type is material inside

rmu75:
fix flange length in case bend type is "material outside". 

relevant discussion: https://forum.freecadweb.org/viewtopic.php?f=3&t=60818&start=570#p588356

### DIFF
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -671,9 +671,11 @@ def smBend(thk, bendR = 1.0, bendA = 90.0, miterA1 = 0.0,miterA2 = 0.0, BendType
   elif BendType == "Material Inside" :
     offset = -(thk + bendR)
     inside = True
+    extLen = extLen + offset
   elif BendType == "Thickness Outside" :
     offset = -bendR
     inside = True
+    extLen = extLen + offset
   elif BendType == "Offset" :
     if offset < 0.0 :
       inside = True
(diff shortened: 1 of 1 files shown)


shaise:
doing that will break old files. meaning if you load a file saved by older version and recompute it will change its dimensions.
We need to ask the developer who added this feature (I think its @jaisekjames) for his opinion on this, was it this his intention?  If its a bug, then it should be fixed even if it breaks old files.

rmu75:
I'm not sure how to fix this properly. In case you design for manufacturing for a press brake, specifying leg length doesn't make much sense in the first place. Maybe an additional field how to interpret the length field can be added that defaults to the behaviour as it is now.

rmu75:
so it seems commercial CAD programs know of "outer virtual sharp", "inner virtual sharp" and "tangent" for flange lengths. the leg length in freecad is neither. I could add a property that defaults to current behaviour for backward compatibility that specifies how this length should be interpreted.

rmu75:
I updated the PR to add a property "LengthSpec" which defaults to current behaviour, old documents should not change dimensions. Inner Sharp and Outer Sharp don't work quite correctly yet.

shaise:
Nice! I like your solution. Now when I use "Material inside" + "Tangetial", changing the radius does not change box dimensions!
Thanks!

PR #237, opened 2022-07-21
SM: Fix smExtrude for cylindrical connecting faces

Roy-043:
The `smExtrude` function did not determine the normal of the connecting face at the connecting edge. This could result in a wrong direction for cylindrical connecting faces.

Forum topic:
https://forum.freecadweb.org/viewtopic.php?f=3&t=70433

Note:
Do not forget to update the version number when merging.


shaise:
Good fix! thanks!

PR #242, opened 2022-07-26
SM: Fix getBendetail and smExtrude for non-planar connecting faces

Roy-043:
An improved version of https://github.com/shaise/FreeCAD_SheetMetal/pull/237 for `smExtrude`. The thkDir is now derived from the planar selFace. This leads to a better result if the connecting faces have an angle with the selFace that is not 90 degrees (although such an angle is probably technically incorrect).

The same solution was also applied to `getBendetail`. This solves the "Make Wall" bug for cylindrical connecting faces that was reported on the Forum: https://forum.freecadweb.org/viewtopic.php?p=612859#p612859

Additionally:
The revAxisV vector is not used in `smExtrude` and was therefore removed.

### DIFF
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -360,10 +360,12 @@ def getBendetail(selFaceNames, MainObject, bendR, bendA, flipped, offset, gap1, 
 
     # main Length Edge
     revAxisV.normalize()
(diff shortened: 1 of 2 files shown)


shaise:
Thanks!

PR #244, opened 2022-07-30
SM: Make PR242 compatible with FreeCAD version 0.20

Roy-043:
This PR fixes the problem introduced with PR242.


shaise:
Thanks for the support!

PR #245, opened 2022-08-03
[Bugfix] Gives more flexibility to use with multiple versions of FreeCAD

Syres916:
See discussion https://forum.freecadweb.org/viewtopic.php?f=3&t=70742


shaise:
Does this work both on 19.4 and 19.3?

Syres916:
Yes

PR #246, opened 2022-08-04
Retain Shape Color and Transparency of base object

Syres916:
See Issue https://github.com/FreeCAD/FreeCAD/issues/7085


shaise:
Hi @Syres916,

Thanks for this important fix!
Hope its ok with you, I have converted all your changes to use functions instead of repeated code, this way if we have a solution to face colors, we can do it in a single place.

Thanks!
shai

PR #248, opened 2022-09-05
Fixed unfolding script when the same edge is defined twice but with vertices in a different order

chappatted:
I could not unfold some parts with holes. The problem is that some CAD softwares will split the hole geometry in 2 surfaces instead of just one, and the unfold script would fail in those cases. I fixed this by implementing a custom edge comparison method, and by cleaning some code that was doing things for no apparent reason.

### DIFF
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -558,7 +558,7 @@ class SheetTree(object):
           for lookVert in self.f_list[i].Vertexes:
             if equal_vertex(lookVert, analyVert):
               if len(theEdge.Vertexes) == 1: # Edge is a circle
-                if not self.is_sheet_edge_face(theEdge, theNode, i):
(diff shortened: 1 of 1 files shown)


shaise:
Hi.
Thanks for the fix!
However, are you sure the following code can really be deleted?
```
    the_index = None
    has_sheet_distance_vertex = False
    for i in self.index_list:
      for sf_edge in self.f_list[i].Edges:
        if sf_edge.isSame(ise_edge):
          the_index = i
          #print 'got edge face: Face', str(i+1)
          break
      if the_index is not None:
        break
```
You have removed this code and sent the index instead. But this code also checks that another face shares the same edge. In your case it will never be None as you send a valid index all the time.

chappatted:
> But this code also checks that another face shares the same edge.

Ok I think I missunderstood that code, I thought it was just searching for the same face again. I put it back.

shaise:
Thank you!
very nice fix!

PR #255, opened 2022-11-07
Improved the unfolding script to handle counterbore holes

chappatted:
Parts with counterbore holes can't be unfolded because the sheet thickness is not constant. See this part for example : [counterbore.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/9950638/counterbore.zip)

Here is a proposition to fix it. When building the sheet tree, the counterbores are detected and saved into a list. When the faces are cloned to build the new unfolded shape, the counterbores are replaced with regular holes. The diameter of the new hole is the smaller radius of the counterbore, since this is what matters in the sketch for laser cutting.

### DIFF
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -1418,7 +1418,7 @@ class SheetTree(object):
       return False
 
     # we use dot product with face normals to check if they are parallel
-    dot = self.face_normal(next_faces[0])[0].dot(self.face_normal(parent_face)[0])
+    dot = self.face_normal(next_faces[0]).dot(self.face_normal(parent_face))
 
     if math.isclose(dot, 1) or math.isclose(dot, -1):
       # since there is an intermediate face parallel to the parent face, this is a counterbore, let's skip this face
@@ -1519,6 +1519,12 @@ class SheetTree(object):
           return i
 
     return None
+
+  # Compute the normal of a face
+  def face_normal(self, face):
+    uv = face.Surface.parameter(face.CenterOfGravity)
+
+    return face.normalAt(uv[0], uv[1])
 
   def searchNode(self, theIdx, sNode):
     # search for a Node with theIdx in sNode.idx
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -407,6 +407,15 @@ def get_surface(face):
   return surface
 
 class SheetTree(object):
+
+  # Class representing a wire to replace in the unfolded shape. During tree creation, some features are detected
+  # (e.g. countersink and counterbore holes) and are replaced later when the unfolded shape is created.
+  class WireReplacement:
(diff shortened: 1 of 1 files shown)


luzpaz:
An aside, it be great to have some unittests that we can test new PRs like this against

shaise:
your example does not unfold with your new code:
```
16:25:29  exception at line 3068("'SheetTree' object has no attribute 'face_normal'",)
16:25:29  ("'SheetTree' object has no attribute 'face_normal'",)
16:25:29  Traceback (most recent call last):
16:25:29    File "C:\Users\shai\AppData\Roaming\FreeCAD\Mod\FreeCAD_SheetMetalcb\.\SheetMetalUnfolder.py", line 3068, in accept
    s, foldComp, norm, thename, err_cd, fSel, obN  = getUnfold(k_factor_lookup)
16:25:29    File "C:\Users\shai\AppData\Roaming\FreeCAD\Mod\FreeCAD_SheetMetalcb\.\SheetMetalUnfolder.py", line 2376, in getUnfold
    TheTree.Bend_analysis(f_number, None) # traverses the shape and builds the tree-structure
16:25:29    File "C:\Users\shai\AppData\Roaming\FreeCAD\Mod\FreeCAD_SheetMetalcb\.\SheetMetalUnfolder.py", line 1357, in Bend_analysis
    if not self.handle_hole(parent_node, face_idx, edge, child_face, child_index):
16:25:29    File "C:\Users\shai\AppData\Roaming\FreeCAD\Mod\FreeCAD_SheetMetalcb\.\SheetMetalUnfolder.py", line 1421, in handle_hole
    dot = self.face_normal(next_faces[0])[0].dot(self.face_normal(parent_face)[0])
16:25:29  AttributeError: 'SheetTree' object has no attribute 'face_normal'
16:25:29  
```
I use freecad 20 release version

chappatted:
Sorry, a method was missing. I added it, it should be working now.

shaise:
Hi, tested, Works!
Usually, this type of part is considered "non legal" sheet metal part, but as long as it works correctly on "legal" parts I'm all for making the unfolder as forgiving as possible. (for example, it would be nice if the unfolder can support countersunk holes as well as counter bore)
So thanks!!!

chappatted:
> Hi, tested, Works! Usually, this type of part is considered "non legal" sheet metal part, but as long as it works correctly on "legal" parts I'm all for making the unfolder as forgiving as possible. (for example, it would be nice if the unfolder can support countersunk holes as well as counter bore) So thanks!!!

Well it may not be "legal" because the sheet thickness is not constant, however having countersink/counterbore in a sheet metal is not necessarily a rarity. Next thing I would like to fix is unfolding parts with a chamfer, and we have the same problem there (thickness is non constant but this can happen in real world parts).

shaise:
I completely agree. thanks!

PR #256, opened 2022-11-16
Fix to countersink detection when face is a cone instead of a cylinder

chappatted:
When a countersink is defined with a cone, the unfolding would fail. This does fix it. Here is an exemple part: [Lamage conique.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/10023167/Lamage.conique.zip)

### DIFF
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -1481,7 +1481,7 @@ class SheetTree(object):
 
   # Check if a face is cylindrical or not.
   def is_cylindrical_face(self, face):
-    return str(get_surface(face)) == "<Cylinder object>"
+    return str(get_surface(face)) == "<Cylinder object>" or str(get_surface(face)) == "<Cone object>"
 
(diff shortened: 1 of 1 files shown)


shaise:
Interesting. For the following part (partly chamfered):
image

when you select the upper face and unlfold, you get cylinders:
image

and when you select the bottom face, unfold is only cones (without the cylinder part):
image

chappatted:
This is actually "normal" because of how the algorithm works:

- If you select the top face, it will detect the counterbore/countersink and replace the circle of the top face because the bottom face has a smaller radius.
- If you select the bottom face, it will just handle it as a normal hole because the circle is smaller than the other one. However, in this case the unfold sketch is generated from the bottom face (and not the top one) so it will still be right.

So in both cases we get a sketch with the right circle (smallest radius). Still, I fixed the unfolded shape for consistency. Please feel free to retest your part.

shaise:
Thanks a lot! this is a great solution.

PR #258, opened 2022-12-22
Improved unfolding script to better handle chamfers

chappatted:
Unfolding parts with chamfers would always fail because the sheet thickness is not constant. I propose to fix it by reconstructing the chamfered faces (approach similar to counterbores and countersinks). Attached two parts that can be unfolded with the modified script.

[parts.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/10285845/parts.zip)

### DIFF
--- a/SheetMetalUnfolder.py
+++ b/SheetMetalUnfolder.py
@@ -1347,7 +1347,7 @@ class SheetTree(object):
           edge = child_info[1]
 
           if not self.handle_hole(parent_node, face_idx, edge, child_face, child_index):
-            if not self.handle_chamfer(face_idx, edge, child_face):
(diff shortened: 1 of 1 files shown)


shaise:
Hi,

Thank you for your work,
However your changes seems to break the unfolder functionality.
The attached file can not be unfolded now.

shai
[SheetMetal_unfold_test.zip](https://github.com/shaise/FreeCAD_SheetMetal/files/10293869/SheetMetal_unfold_test.zip)

chappatted:
Hi,

There was indeed a problem with that part, the script detected some of the foils as chamfers. I added an extra check to avoid that problem. Please feel free to test it again.

shaise:
Thanks! looks good now.

PR #263, opened 2023-01-24
Add the option to auto-link wall bend radius

### LINKED ISSUE #257: [Enhancement] Fold radius
Hello.

Right now if we fold a wall, we have to change fold radius for each fold. It's tedious if we make a lot of separate folds.
What do you think about making radius value of first fold the main value, which would be inherited by every next fold + add another property which will allow to override this value?

Sorry for poor english :)

alexneufeld:
Adds a new preferences entry to automatically link the bend radius of newly created sheet metal walls to their earliest dependent Bend or BaseBend object.

Closes #257

image

image

This is a first attempt, I would appreciate comments on the implementation and design of this feature, as well as testing to see if it breaks in any edge cases.

### DIFF
--- a/SMprefs.ui
+++ b/SMprefs.ui
@@ -63,10 +63,66 @@
            <number>0</number>
           </property>
           <property name="prefEntry" stdset="0">
-           <string>EngineeringUXMode</string>
+           <cstring>EngineeringUXMode</cstring>
           </property>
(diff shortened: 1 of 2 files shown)


shaise:
Great solution!
I have taken the liberty to implement it over all bend commands, Thank you!

PR #267, opened 2023-03-20
Add translation support

riqueenz:
Support for translation has been added. The texts had the TRANSLATION_NOOP function which made my work a lot easier :-). So basically I just added the translations folder, copied the fasteners workbench helpers and added the translations path at the beginning of the files that have strings.


shaise:
Great work! Thanks!!!

PR #268, opened 2023-03-23
Unfolder DXF/SVG Exporter and improvements..

Syres916:
... to stability with different models since the changes at the end 2022. I've had to keep 0.2.59 available in order to be more confident of unfolding working.

Note: the DXF/SVG files are saved in the same folder as the original FreeCAD file.


shaise:
Thanks!!

Issue #269, opened 2023-03-24
Unable to unfold


olivierCHAM:
The attached design can not be unfold with sheet metal 0.2.61

The attached file includes an unfold done with sheet metal 0.2.59

OS: Mageia 7 (WindowMaker/WindowMaker)
Word size of FreeCAD: 64-bit
Version: 0.20.2.29603 (Git) AppImage
Build type: Release
Branch: (HEAD detached at 0.20.2)
Hash: 930dd9a76203a3260b1e6256c70c1c3cad8c5cb8
Python 3.10.8, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: French/France (fr_FR)
Installed mods: 
  * fasteners 0.4.54
  * A2plus 0.4.60i
  * workfeature-macro
  * sheetmetal 0.2.61
[C520-210-pliage.FCStd.gz](https://github.com/shaise/FreeCAD_SheetMetal/files/11060229/C520-210-pliage.FCStd.gz)

Syres916:
As I mentioned in the forum, please update to 0.2.62 and retry, thanks @shaise for merging so quickly.

olivierCHAM:
I have updated to 0.2.62 and the unfold is OK
Thank you all for your response

GS90:
Version: 0.2.62
Unstable operation "Unfold"...

![sm](https://user-images.githubusercontent.com/25671299/227706841-9d509c1c-f315-4f59-b464-56c3fa29291f.jpg)

error:
**Unfold is failing.
Please try to select a different face to unfold your object.**
```
Engineering mode is not enabled.
Manual K-factor is being used: 0.50 (ansi)
exception at line 3346("'FeaturePython' object has no attribute 'Refine'",)
("'FeaturePython' object has no attribute 'Refine'",)
Traceback (most recent call last):
  File "/home/***/.local/share/FreeCAD/Mod/sheetmetal/./SheetMetalUnfolder.py", line 3346, in accept
    s, foldComp, norm, thename, err_cd, fSel, obN  = getUnfold(k_factor_lookup)
  File "/home/***/.local/share/FreeCAD/Mod/sheetmetal/./SheetMetalUnfolder.py", line 2626, in getUnfold
    TheTree.Bend_analysis(f_number, None) # traverses the shape and builds the tree-structure
[...]

PR #270, opened 2023-03-25
Fix newly introduced bugs for simple models

Syres916:
Retested on both complex and simple models. Added extra help for users in the error message box regarding the use of Refine on features. See report from user GS90 https://github.com/shaise/FreeCAD_SheetMetal/issues/269


shaise:
Thanks!

PR #275, opened 2023-04-06
Remind user that the bend sketch ...

Syres916:
... should be of sufficient length


Syres916:
Gentle and polite nudge, @shaise could you just let know when you'll be able to review this please, TIA.

shaise:
Wow, sorry about that... but better late then never.

PR #288, opened 2023-12-09
Update of the translations in Polish

kaktusus:
A file update has been performed:
`SheetMetal_pl.qm`
`SheetMetal_pl.ts`
for the Polish language.


shaise:
thanks!

kaktusus:
I would like to ask you to notify the translators when an update is due. Not everyone pays attention if the changes affect the GUI.

If it's not a problem, could you post information about the update on the FC forum in the [translators section](https://forum.freecad.org/viewforum.php?style=4&f=45)  ?

shaise:
Hi @kaktusus ,
The problem is there are many small changes over time, and its not feasible to merge lots of pull requests after every small change. but I will try to notify each time there is a major change.

PR #293, opened 2024-01-16
[Sheetmetal] Python 3.11.x and above Json Fix

Syres916:
In order to fix errors using Python 3.11.x and above while preserving ability for Python 3.10.x and below to still work correctly, example error:

```
  File "/usr/lib/python3.11/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
<class 'TypeError'>: Object of type FeaturePython is not JSON serializable
17:04:02  PropertyPythonObject::toString(): failed for <class 'SheetMetalCmd.SMViewProviderFlat'>
17:04:02  pyException: Traceback (most recent call last):
  File "/usr/lib/python3.11/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
```

@shaise Certainly worth testing with a Python 3.10 or below to ensure there's no regressions but I'm quite happy this should be good to go. Any questions, let me know.

### DIFF
--- a/SheetMetalBaseCmd.py
+++ b/SheetMetalBaseCmd.py
@@ -284,6 +284,17 @@ class SMBaseViewProvider:
             doc = FreeCAD.ActiveDocument  # crap
             self.Object = doc.getObject(state["ObjectName"])
 
+    # dumps and loads replace __getstate__ and __setstate__ post v. 0.21.2
+    def dumps(self):
+        return None
+
+    def loads(self, state):
+        if state is not None:
+            import FreeCAD
+
+            doc = FreeCAD.ActiveDocument  # crap
+            self.Object = doc.getObject(state["ObjectName"])
+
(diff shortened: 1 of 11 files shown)


shaise:
Thanks!

I changes your solution a little bit, so instead of duplicating the code of __setstate__ and loads, I just call loads from __setstate__,

PR #299, opened 2024-02-09
Make license match SPDX ID

chennes:
See https://spdx.org/licenses/ -- note that you might actually mean `LGPL-2.1-only`, in which case use that instead.


shaise:
Thanks!

PR #301, opened 2024-03-01
Update translation script

hasecilu:
Same as in Fasteners WB, with some changes.

I tested both scripts and the generated files is the same.
<details>
    <summary>Click to Déjà vu</summary>
The old script worked good enough but it was hard to read and also didn't follow the workflow (updated and released at the same time when in reality the changes of the translator should be in the middle, minor thing).

The new script:
- is based on the script available at freecad.workbench_starterkit template
- is well documented and easy to read
- files are saved in lowercase
- an language agnostic file is updated every time another file is updated

The script updates `fasteners.ts` file every time any other translation file is created or updated. This is made with the objective of having this file updated with the most current changes. If a translator doesn't have access to `pylupdate` command it can start translating based on the `fasteners.ts` file, it only needs to replace
```xml
<TS version="2.1">
```

for

```xml
<TS version="2.1" language="ja" sourcelanguage="en">
```

using their language code (must be supported on FreeCAD).
    
</details>

- **The difference** with the script on Fasteners WB is that here I'm using `pylupdate6`.
Originally I wanted to use `pylupdate6` on Fasteners but I got an error complaining 
```
pylupdate6: Invalid syntax at line 1 of ../CountersunkHoles.py:
# -*- coding: utf-8 -*-
```
Fortunately this works here, not really sure why because some file also have `# -*- coding: utf-8 -*-`

The advantage to use `pylupdate6` over `pylupdate5` is that some strings that are formatted are not picked with `pylupdate5` but they are with `pylupdate6`, `pylupdate5` is buggy, in fact, the first time I translated this WB even though I finished all lines some commands were missing, `pylupdate5` was responsible. Example: the comma after the string in Python is recommended in case you want to add another element to the list but `pylupdate5` fails to catch it.
```python
ToolTip = FreeCAD.Qt.translate(
        "SheetMetal",
        "Sheet Metal workbench allows for designing and unfolding sheet metal parts",
    )
```

- Another thing, when running `	lupdate ../*.ui -ts uifiles.ts -no-obsolete`  it causes an error because the file `lookup.ui` is not a proper XML file, is it really neeeded or can be deleted? Aternatively I can replace the wildcard (`*`) to make it ignore that file.

- I made the name of the WB translatable as in Fasteners WB
**Edit:** make translatable menu, toolbar and context menu (right click)


shaise:
Thanks!!

PR #310, opened 2024-03-17
More explicit workbench icon

pierreporte:
I propose new icons that show what kind of part is going to be made with the workbench, instead of some “construction work” pattern. The base shape is done with FreeCAD and the SheetMetal worbench, then colored in Inkscape.


shaise:
It's a bit funny: 2 years ago or so, I changed the icon to something like you proposed, and I got rejects by the community, so I changed it back...

pierreporte:
What were the arguments? You can hardly think of sheet metal when seeing the current icon for the first time.

shaise:
Yes, I tend to agree. Specially since there is a talk to integrate this WB to core freecad.
I need to check with someone, and if he agrees, I will merge it.

MisterMakerNL:
> It's a bit funny: 2 years ago or so, I changed the icon to something like you proposed, and I got rejects by the community, so I changed it back...

Things change :)
I support this PR.

maxwxyz:
I've modified it a bit to follow https://wiki.freecad.org/Artwork_Guidelines
Hope its OK, you can use this .svg if you like.
Will redo the other icons next week to be consistent.
preferences-sheetmetal

pierreporte:
Thank you @maxwxyz for the updated icon.

obelisk79:
I like this icon

shaise:
I like it too! 
Thank you @maxwxyz and @pierreporte !
Lets make a change.

PR #314, opened 2024-03-21
Rename commands to coincide with Wiki pages

hasecilu:
Same as in Fasteners WB, to be able to use the `What's this?` command we need to synchronize the names.

You need to choose what names you prefer to be used.

| WB name | Wiki name |    |
|---------|-----------|----|
|Base| AddBase|  |
| MakeWall | AddWall |  |
| ExtrudeFace | Extrude |  |
| FoldWall | AddFoldWall |  |
| Unfold | Unfold | OK |
| UnfoldUnattended | UnattendedUnfold |  |
| CornerRelief | AddCornerRelief |  |
| MakeRelief | AddRelief |  | 
| MakeJunction | AddJunction |  |
| MakeBend | AddBend |  |
| SketchOnSheet | SketchOnSheet | OK | 
| FormingWall | Forming | |
| BaseShape | BaseShape | OK |

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -43,7 +43,6 @@ main_smWB_Icon = os.path.join(smWB_icons_path, "SMLogo.svg")
 
 
 class SMWorkbench(Workbench):
-
     global main_smWB_Icon
     global SHEETMETALWB_VERSION
     global QtCore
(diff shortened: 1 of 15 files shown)


shaise:
The problem is, I think, If the command is a creation command, changing it might break old saved files. We need to test it.

shaise:
Ok, I did some checks. The saved files does not hold the command name, they hold the base class of the command. So it probably not hurt old files. It might hurt old macros, but I think this is not a big issue.

hasecilu:
> Ok, I did some checks. The saved files does not hold the command name, they hold the base class of the command. So it probably not hurt old files. It might hurt old macros, but I think this is not a big issue.

Yes, I proposed similar changes on Gears WB and my tests show that the objects only save the class that created them, and as long that doesn't change there is no problems.
https://github.com/looooo/freecad.gears/pull/153

hasecilu:
Now the Wiki pages are being open correctly.

shaise:
Thanks!!

PR #319, opened 2024-03-29
Some refactoring on bendsolid.

sliptonic:
Some Black() formatting.  Renamed some variables to make the code more readable.
Some minor refactoring.

Added unit test structure and a unit test for get_point_on_cylinder

### DIFF
--- a/SheetMetalBendSolid.py
+++ b/SheetMetalBendSolid.py
@@ -25,116 +25,222 @@
 
 import Part, math
 
-def getPointOnCylinder(zeroVert, poi, Radius, circent, axis, zeroVertNormal):
-    dist = zeroVert.distanceToPlane(poi, zeroVertNormal)
-    poi1 = poi.projectToPlane(zeroVert, zeroVertNormal)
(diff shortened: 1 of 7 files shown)


shaise:
Hi @sliptonic 

It looks like there are some conflicts.
I recently made quite a lot of bug fixing and algorithm changes to sheetmetal, so this might cause the conflicts.

shai

sliptonic:
Ok.  Rebased and cleaned up.  
This is mostly minor stuff.  variable naming, black(), and a unit test.  
Everything works with my test models but could use a double check.

shaise:
Ok I will take a look. I have added more functionality, this makes bend function work on refined faces as well, and bend solid automatically calculates the thickness. I should add these models as a test case.

shaise:
Ok. It looks like it works on the new test models as well.
I see that you've change the function names and some variables to ["Snake Case" ](https://pythonguides.com/python-naming-conventions/)
Is there a formal naming conversion I should follow?
shai

PR #333, opened 2024-04-25
Discard unneeded int casts in getStandardButtons methods

YakoYakoYokuYoku:
Because `QtWidgets.QDialogButtonBox.StandardButtons` are already converted into `int`s in FreeCAD's `Gui::TaskDialogPython::getStandardButtons`, such conversions in workbench code aren't needed. Also this fixes compatibility with PySide6.

https://github.com/FreeCAD/FreeCAD/blob/b9bfa5c5507506e4515816414cd27f4851d00489/src/Gui/TaskView/TaskDialogPython.cpp#L736-L754

### DIFF
--- a/SheetMetalBend.py
+++ b/SheetMetalBend.py
@@ -347,7 +347,7 @@ class SMBendTaskPanel:
         return True
 
     def getStandardButtons(self):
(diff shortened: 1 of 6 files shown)


shaise:
Thanks!

PR #338, opened 2024-05-21
Make parameter `shapeType` enum translatable

hasecilu:
- Strings now also appear translated in the Property View, the same as in the dockable UI
- Fix hard-coded strings on `updateEnableState()` method
image

### DIFF
--- a/SheetMetalBaseShapeCmd.py
+++ b/SheetMetalBaseShapeCmd.py
@@ -38,12 +38,20 @@ mw = FreeCADGui.getMainWindow()
 # changes in modeling logic
 smElementMapVersion = 'sm1.'
 
-base_shape_types = ["Flat", "L-Shape", "U-Shape", "Tub", "Hat", "Box"]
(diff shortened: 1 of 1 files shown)


shaise:
Hi @hasecilu ,

Sorry, this will not work. Since your translation goes all the way down to the command property as well, loading a file saved in another language will cause an error
Perhaps we need to translate only the gui. 

shai

hasecilu:
Hi, I made several tests and it seems is like that, previously only tested on current session.

Do you know if FreeCAD has a method to get the **index** of the enum property? When we retrieve the value it returns the string, i think would to the trick.

shaise:
When freecad saves an enum parameter, it saves it name. When it reads it, it must be one of the enums it have in the list.

hasecilu:
Well, for now I think the method `OnChanged()` could be made using the strings, I'll make another PR

shaise:
As said, the gui (task window)  can still be translatable - but then set the property with the original values.

PR #339, opened 2024-05-22
Make unused parameters `readonly` when are not used

hasecilu:
Same behaviour as in the dockable UI

Example when "Flat" can't edit radius.
image


shaise:
Thanks!

Issue #350, opened 2024-06-27
Basename of the Unfolded objects


leoheck:
Hi @shaise , I would like to suggest to use the base Body name (actually Label) as the basename of the unfolded objects.

For instance, the object name is `Bottom`
Then if I unfold it, it could generate the `Bottom_Unfolded` and the `Bottom_Unfolded_Sketch`

![image](https://github.com/shaise/FreeCAD_SheetMetal/assets/1277920/6f2dfa35-840f-49f5-975c-a7990c65c1ec)

shaise:
Sounds like a good idea.
Added to my to do list

leoheck:
Nice, thanks. I tried to improve this here https://github.com/shaise/FreeCAD_SheetMetal/pull/352. What do you think?

leoheck:
The proposal works as you can see in the screenshot below.

I just noticed an interesting thing.
I have enabled the Freecad option to allow object with the same Label.
So no more `001, 002, 003..` at the end of repeated objects.

However when I repeat the unfolding of the same object.
It creates this new Part ending with `001` and a new Sketch with the right name as in my configs. I am not sure if this is an issue with the Sheetmetal or if this is an issue with my latest 0.22-dev build.

What do you think?

![image](https://github.com/shaise/FreeCAD_SheetMetal/assets/1277920/05c24e88-d347-411c-a26c-1cd1363ca3a6)

leoheck:
Ah nevermind, this looks like a bug on the latest dev.

shaise:
of object names are not unique anymore how can you tell them apart?

leoheck:
hm did not know, I got the first parent from the Parents list and on my test, it worked. maybe I got lucky.

leoheck:
Actually all partents are the same, maybe...

This example here:
![image](https://github.com/shaise/FreeCAD_SheetMetal/assets/1277920/e7ab5f9a-d7a2-4bae-9b7a-d01b3a11e795)


```python
Gui.Selection.addSelection('Unnamed','Body001','Bend001.')
>>> ### Begin command Std_SendToPythonConsole
>>> # try:
>>> #     del(doc,lnk,obj,shp,sub,subs)
>>> # except Exception:
>>> #     pass
>>> # 
>>> # doc = App.getDocument("Unnamed")
>>> # obj = doc.getObject("Bend001")
>>> # shp = obj.Shape
>>> ### End command Std_SendToPythonConsole
>>> obj.Label
[...]

PR #352, opened 2024-06-27
Improve the objects Label of unfolded objects. 

leoheck:
This is a try to improve unfolded object names, https://github.com/shaise/FreeCAD_SheetMetal/issues/350

### DIFF
--- a/SheetMetalNewUnfolder.py
+++ b/SheetMetalNewUnfolder.py
@@ -1353,6 +1353,7 @@ def getUnfold(
 
 
 def getUnfoldSketches(
+    obj_label,
     selected_face: Part.Face,
     unfolded_shape: Part.Shape,
     bend_lines: Part.Compound,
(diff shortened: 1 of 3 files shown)


luzpaz:
soft bump

leoheck:
I updated the code @luzpaz 
Could you please test it?

luzpaz:
>  `SMLogger.error(FreeCAD.Qt.translate("Logger", f"tidying up {unfold_base_label}_Sketch_Outline"))`

@leoheck I'm not sure the translation function will work as you hope here.

leoheck:
hm yeah, thanks. The update should fix it.

leoheck:
@luzpaz ^

luzpaz:
@shaise can you review ?

shaise:
@leoheck , @luzpaz 

There are several issues:
1. the workbench does not load at all. there are several syntax errors.
2. the renaming work only on the old unfolder, not on the new one which is the default.
3. there are double "unfold" in the names:
image

This should be only the base name without adding an extra "Unfolded".
I would change the code:
```
    try:
        unfold_base_label = obj_label + "_Unfolded"
    except:
        unfold_base_label = "Unfolded"
```
to:
```
    if obj_label is None:
        obj_label = "Unfolded"
```

and use obj_label instead of unfold_base_label.
so the result:
image

leoheck:
Thanks for your suggestions @shaise .
I fixed the syntax in v1 and also added the same behavior to the unfolder v2.

PR #358, opened 2024-07-13
Move tests to unique folder name due to a clash with the CAM...

Syres916:
...Wb Tests. the Sheetmetal Tests folder is currently picked up in the path before CAM and so no CAM tests run. As CAM is a core workbench it should get priority, though it too should have a unique name IMHO.


shaise:
thanks!

PR #359, opened 2024-07-17
[SheetMetalUnfolder] Sketch colors must be integers...

Syres916:
...not floats. Assumption: the new Materials system has changed this so I made sure it still works with 0.21.x and it does.

Currently my box puts out loads of `QColor::setRgb: RGB parameters out of range` warnings, this is fixed with the PR.


shaise:
Thanks!!

PR #362, opened 2024-08-19
Fix lost of topo naming

realthunder:
Related realthunder/FreeCAD#1013. The root cause is lost of topo naming due a bug in `mapShapes()` and is fixed in [here](https://github.com/realthunder/FreeCAD/commit/aa1aa00ae4784ebfae85407e8bdc932c456858cf). There is also a minor problem in your code, hence the PR here.

In my branch, you can check if topo naming is working by mouse over highlight. It should show the added shape.

Screenshot from 2024-08-19 20-58-53


shaise:
Thanks a lot! you are my hero!

PR #382, opened 2024-10-12
[SheetMetalUnfolder] Compatibility with FreeCAD < 0.21

GS90:
Hi @shaise.
Unfortunately, due to incomplete backward compatibility, we have to use older versions of FreeCAD.


shaise:
Thanks!

PR #396, opened 2024-11-13
Fix and improvement to Extruded Cutout: related to Cut type "Through everything" and getting the user selected sketch

sheetmetalman:
The fix is in Cut type "Through everything":
- Before this fix the code get the bound box diagonal of the sheet metal part. But when the sketch far away from the sheet metal part (like the image below) the "Through everything" didn't work.
- But with the fix, the code gets the distance from the sketch to the sheet metal part plus the bound box diagonal. This ensures that it goes through everything.

The improvement is in the code to get the sketch from user selection, now this part of the code is smaller.

image


shaise:
Thanks!

PR #397, opened 2024-11-16
Update ExtrudedCutout.py for correcting the creation of faces from the sketch (now supports multiple cuts from a single sketch)

sheetmetalman:
Before this fix, Extruded Cutout feature cannot create a cut when the wires from a single sketch creates more than one separate face.

Now, with this fix, a single sketch can have wires that create more than one separate face, thus allowing the creation of multiple cuts from a single sketch.

image


shaise:
great! thanks!

Issue #402, opened 2024-11-21
Error when loading a file with extruded cutout features


shaise:
### Describe your issue

Loading a file with extruded cutout features generates some errors, even though the file seems ok after it is loaded.
I was able to remove some of the errors (missing loads() function) but others still exist
To reproduce, just open the attached file.

### FreeCAD version info + SheetMetal WB version

```shell
OS: Windows 11 build 22631
Architecture: x86_64
Version: 1.1.0dev.39100 (Git)
Build type: Release
Branch: main
Hash: 8865450a3e14220925e0e449c0f1f79056b4fb89
Python 3.11.10, Qt 5.15.15, Coin 4.0.3, Vtk 9.3.0, OCC 7.8.1
Locale: English/United States (en_US)
Stylesheet/Theme/QtStyle: OpenLight.qss/OpenLight/Fusion
Installed mods: 
  * fasteners 0.5.31
  * OpenTheme 2024.9.1
  * sheetmetal 0.5.6
  * STEMFIE 0.2.4
```

### Put here any screenshots or videos (optional)

[test-cutout2.zip](https://github.com/user-attachments/files/17845980/test-cutout2.zip)

shaise:
@sheetmetalman , I'm still looking for the problem. Please update your code as I fixed some of it.

shaise:
Ok I fixed it. Please update.
What was the problem:
there is a function "onChanged" in class ExtrudedCutout. This calls recompute each time a value is changed.
Problem is during initialization when the fuction is called EACH time addProperty is called, causing the system to recompute even though other parameters do not exist yet...
solution: Remove the recompute. it is done automatically anyway.

sheetmetalman:
Thank you so much, @shaise!

Yes! I will update my code here same as yours updates! Tanks again!

Issue #401, opened 2024-11-20
       Extruded Cutout    Incorrect work


free777cat:
### Describe your issue

     Extruded Cutout    Incorrect work

### FreeCAD version info + SheetMetal WB version

```shell
[code]
OS: Linux Mint 22 (MATE/mate)
Word size of FreeCAD: 64-bit
Version: 2024.1006.0.14555 (Git shallow) AppImage
Build type: Release
Branch: (HEAD detached at 20241003stable)
Hash: 8058dcf52c2661edabf36d8147d5b8a187378d2d
Python 3.11.9, Qt 5.15.13, Coin 4.0.1, Vtk 9.2.6, OCC 7.7.2
Locale: Ukrainian/Ukraine (uk_UA)
Installed mods: 


  * sheetmetal 0.5.6

[/code]
```

### Put here any screenshots or videos (optional)

[Incorrect_work.zip](https://github.com/user-attachments/files/17827539/Incorrect_work.zip)
![Зняток екрана як 2024-11-20 10-39-43](https://github.com/user-attachments/assets/1a591179-20e7-412e-92e2-b0bf73e75a5b)

shaise:
@sheetmetalman , can you take a look?

sheetmetalman:
Yes! As soon as possible.

sheetmetalman:
Hi!

As I understood, the problem is that the result geometry is entering the sketch extrusion zone, as show in the image below.
But the correct geometry should be the second image.

![image](https://github.com/user-attachments/assets/5bca4a5a-3f7a-42da-8ed0-5754314ae50e)
![image](https://github.com/user-attachments/assets/ac368387-3340-4c2d-b53a-e5a9eef19df0)

Looking other software (SolidWorks) at how they deal with, I find they have the same issue, BUT they implement a function to resolve this. Solidworks calls it "Optimize geometry". The below images and video show what happen in this case in SolidWorks.

![image](https://github.com/user-attachments/assets/8104809d-cfa7-4891-b74f-be5f437e7b11)
![image](https://github.com/user-attachments/assets/c07963c1-c0c7-485e-8334-80fe0d9a49a4)
![image](https://github.com/user-attachments/assets/c8af1851-33b1-4a78-a912-4d183697138c)
![image](https://github.com/user-attachments/assets/bac49cee-7297-443b-aa0e-1dc5a1ea72b7)

[...]

PR #403, opened 2024-11-22
Extruded Cutout: add 2 properties "Improve cut" and "Improve Level"

sheetmetalman:
Related to issue #401.

The property "Improve cut" is way to solve the problem of the result geometry is entering the sketch extrusion zone.

But it's not perfect yet, it's not solve 100% of the problem. The evidence of this is the new property "Improve Level". High number of this property approach to the 100% solution. I consider number 4 to be sufficient for most cases.

When this property is activated the speed performance of the feature decreases considerably. Because of perfomance I have put limits on the value of the "ImprovementLevel" property: maximum 20 (even this value takes a very long time to compute, more than this will lead too much time on computing, will low too much the speed of the computing). The minimum of 2 is because 1 causes error.

The next step is to find a way to smooth the edges before the cut operation in the code.

I also: put the "refine" property and theses 2 news properties into a new property category called "ExtrudedCutout Improvements". And removed that debugging lines from issue #402 and other fixes relateds (functions "dumps" and "loads" in the classes of view providers).

@shaise. I was thinking if is best to find a way to smooth the edges before merge this code. But maybe is best merge this code now and when the "smooth" is done, merge it. What do you think?

========================================================================
There's two advices on the gui to user about using this property.

Abouth the "Improve Cut": "Improve cut geometry if it enters the cutting zone. Only select true if the cut needs fix, 'cause it can be slow".

About the "Improve Level": "Level of cut improvement quality. More than 10 can take a very long time".

https://github.com/user-attachments/assets/5995159a-d35c-4afd-9878-d56e4e2b3343

### DIFF
--- a/ExtrudedCutout.py
+++ b/ExtrudedCutout.py
@@ -13,11 +13,12 @@ class ExtrudedCutout:
         obj.addProperty("App::PropertyLink", "Sketch", "ExtrudedCutout", "The sketch for the cut").Sketch = sketch
         obj.addProperty("App::PropertyLinkSub", "SelectedFace", "ExtrudedCutout", "The selecteds object and face").SelectedFace = selected_face
         self._addProperties(obj) # Add other properties (is necessary this way to not cause errors on old files)
+        obj.setEditorMode("ImproveLevel",2) # Hide by default
         obj.addProperty("App::PropertyLength", "ExtrusionLength1", "ExtrudedCutout", "Length of the extrusion direction 1").ExtrusionLength1 = 500.0
         obj.setEditorMode("ExtrusionLength1",2) # Hide by default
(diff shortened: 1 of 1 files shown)


sheetmetalman:
The way this works is simple. It simply creates more surfaces between each side of the sheet, and intersects with the sketch extrusion. The higher the value of the "PropertyLevel" property, the greater the number of intermediate surfaces. Then, each surface is offset.

The images below will help explain.
issue 001 01
issue 001 02
issue 001 03
image

shaise:
Hi @sheetmetalman 
Thanks for all that work!
I have made a small change - I reversed the order of the creation of ImproveLevel and ImproveCut properties because otherwise it causes an error when loading old file and recomputing.
I also bumped the version. Please update.

PR #404, opened 2024-11-22
Extruded cutout: remove duplicate lines

sheetmetalman:


shaise:
@sheetmetalman , I have updated the last commit with this fix, so it should be ok now

PR #407, opened 2024-11-25
Update translations from CrowdIn

hasecilu:
This is the most translated FreeCAD WB! :tada: 

Next language is Swedish with "only" 34 string, I leaved it out, should be included? there are even more languages with less strings.

I'll make tests to automate this in the future.

Filename | Translated Messages
-- | --
SheetMetal_es-ar.ts | 250
SheetMetal_es-es.ts | 250
SheetMetal_fr.ts | 250
SheetMetal_sr-cs.ts | 243
SheetMetal_sr.ts | 243
SheetMetal_pl.ts | 233
SheetMetal_de.ts | 231
SheetMetal_it.ts | 186
SheetMetal_el.ts | 176
SheetMetal_zh-TW.ts | 160
SheetMetal_pt-br.ts | 111
SheetMetal_pt-pt.ts | 111


shaise:
Great! Thanks!
Yes, 34 sounds quite low. 
Question is, if its not included, will it discourage translators from continuing? Dilemma.

hasecilu:
> will it discourage translators from continuing? Dilemma.

I really hope it doesn't. Main FreeCAD ask your language to be 25% to be included, even if after a while it gets lower doesn't seems to be removed.

shaise:
> Main FreeCAD ask your language to be 25% to be included.

If there is a freecad policy about that then its good enough!

shaise:
@hasecilu , I think the task ui is not being translated. Do you happen to know what do I need to add to the code to apply the translations?

hasecilu:
Which specific files? If a change was done to the strings that were translated there will be no match and thus no translation.

Edit: Ok, I got some.

shaise:
perhaps I'm wrong. because the base shape ui looks to be translated:
image

Is this done automatically when I load the UI?

hasecilu:
Add Wall panels is not. It's from `FlangeParameters.ui` file which have

```xml
 <class>FlangeParameters</class>
 <widget class="QDialog" name="FlangeParameters">
```

but translation files have strings under different context "SMFlangeTaskPanel"


```xml
  <context>
    <name>SMFlangeTaskPanel</name>
    <message>
      <location filename="../panels/FlangeParameters.ui" line="14"/>
      <source>Flange Parameters</source>
      <translation>Parámetros de brida</translation>
    </message>
```




image

hasecilu:
We can only rename the UI files to match the context on the translations files or we can leave them but update translations files again and recompile to adapt to new context.

image

hasecilu:
> perhaps I'm wrong. because the base shape ui looks to be translated: 
> 
> Is this done automatically when I load the UI?

btw, did baseSHape also starts the base shape selection on "Plain" but the firstly showed shape is L-shape on your end?

shaise:
> btw, did baseSHape also starts the base shape selection on "Plain" but the firstly showed shape is L-shape on your end?

Yes. I think that plane was added in a later time

shaise:
> We can only rename the UI files to match the context on the translations files or we can leave them but update translations files again and recompile to adapt to new context.

I'm OK with both options. Which one is less disruptive?

hasecilu:
Modifying the UI files.

shaise:
OK. is it enough to change the class name within the file or should I change the file name as well?

hasecilu:
No need to rename files, class is enough
```xml
 <class>SMFlangeTaskPanel</class>
```

even there is no need to change `widget` but I always have done both lines
```xml
 <widget class="QDialog" name="FlangeParameters">
```

shaise:
Thanks. I'll get to it asap

shaise:
@hasecilu , 
I have looked through all panels and only one need fixing was "FlangeParameters"
However I have a small issue:
the Flange task panel was once created from 2 ui files: FlangeParameters.ui and FlangeAdvancedParameters.ui
but now everything is under FlangeParameters.ui .
How can we merge them without loosing all the existing translations?
Also, the translation file should be regenerated now after renaming FlangeParameters"

PR #417, opened 2024-12-04
Typos

hasecilu:
There we some issues found on CrowdIn.

---

After running
```shell
 awk 'BEGIN { RS="</source>" } /<source>/ { gsub(/.*<source>/, ""); print }' SheetMetal.ts | aspell --lang=en list | sort | uniq
```
only this strings appear as typos. Some are becaus eof Qt so no way to change it but should we rename/capitalize those sheetmetal strings?

- DXF
- MessageBox
- SVG
- SheetMetal
- Sheetmetal
- SubElement
- UX
- apos
- discretizing
- href
- li
- lt
- ol
- readonly
- sheetmetal

Since 1.0 is close I think it's good idea to check if strings are appropiate/understandable, maybeask users on the forum?

### DIFF
--- a/SheetMetalBaseCmd.py
+++ b/SheetMetalBaseCmd.py
@@ -118,7 +118,7 @@ class SMBaseBend:
         obj.addProperty(
             "App::PropertyLength", "thickness", "Parameters", _tip_
         ).thickness = 1.0
-        _tip_ = translate("App::Property", "Relief Type")
+        _tip_ = translate("App::Property", "Bend Plane")
         obj.addProperty(
             "App::PropertyEnumeration", "BendSide", "Parameters", _tip_
         ).BendSide = ["Outside", "Inside", "Middle"]
@@ -244,7 +244,7 @@ if SheetMetalTools.isGuiLoaded():
                 "ToolTip": translate(
                     "SheetMetal",
                     "Create a sheetmetal wall from a sketch\n"
-                    "1. Select a Skech to create bends with walls.\n"
+                    "1. Select a Sketch to create bends with walls.\n"
                     "2. Use Property editor to modify other parameters",
                 ),
             }
--- a/SheetMetalFormingCmd.py
+++ b/SheetMetalFormingCmd.py
@@ -162,7 +162,9 @@ class SMBendWall:
         obj.addProperty("App::PropertyLinkSub", "toolObject", "Parameters",
                         _tip_).toolObject = (seltool, seltool_items)
         _tip_ = FreeCAD.Qt.translate(
-            "App::Property", "Point Sketch on Sheetmetal")
+            "App::Property",
+            "Sketch containing circle's points to multiply and pattern the embossed feature",
+        )
         obj.addProperty("App::PropertyLink", "Sketch", "Parameters1", _tip_)
         obj.Proxy = self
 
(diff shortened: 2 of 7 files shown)


SheetMetalFormingCmd.py
```diff
@@ -162,7 +162,9 @@ def __init__(self, obj, selobj, selobj_items, seltool, seltool_items):
         obj.addProperty("App::PropertyLinkSub", "toolObject", "Parameters",
                         _tip_).toolObject = (seltool, seltool_items)
         _tip_ = FreeCAD.Qt.translate(
-            "App::Property", "Point Sketch on Sheetmetal")
+            "App::Property", "Point Sketch on Sheetmetal"
+        )  # FIXME: hard to understand
+        # "Sketch containing circle's points to multiply and pattern the embossed feature"
```
hasecilu: This was hard to understand the meaning, only after RTFW (Reading The FreeCAD Wiki) got it, here is proposed change

SheetMetalBaseCmd.py
```diff
@@ -118,7 +118,7 @@ def __init__(self, obj, sketch):
         obj.addProperty(
             "App::PropertyLength", "thickness", "Parameters", _tip_
         ).thickness = 1.0
-        _tip_ = translate("App::Property", "Relief Type")
```
hasecilu, line 121: We believe that "Relief Type" is wrong, it's correct?

InitGui.py
```diff
@@ -50,6 +50,7 @@ class SMWorkbench(Workbench):
     global smWBpath
     global smWB_icons_path
 
+    # TODO: "SheetMetal" or "Sheet Metal"
```
hasecilu: Should the WB name be join or separated?

SheetMetalFormingCmd.py
shaise: The forming tool is quite cumbersome. This is why the UI is not yet implemented. Still thinking how to tackle this.

SheetMetalBaseCmd.py
shaise: yes. I believe It should be 'Bend Plane' or something like this.
Also, does the script understand the 'translate' command, or shall I replace if with the longer 'FreeCAD.Qt.Translate'?

InitGui.py
shaise: I have no preference....

hasecilu: I'll leave as is

SheetMetalBaseCmd.py
hasecilu: I made the change, `translate()` is ok

shaise: sorry if you missed my edit, it should be  'Bend Plane' not wall thickness...

SheetMetalFormingCmd.py
hasecilu: As long as the feature is well documented and working as mentioned I think there is no much problem, should I make the change?

SheetMetalBaseCmd.py
hasecilu: ok

SheetMetalFormingCmd.py
shaise: Yes, thank you.

hasecilu:
Any other change?

shaise:
Looks good so far. 
I have 2 more task panels to implement, but this will take some time.

PR #418, opened 2024-12-05
Fix context from old strings

hasecilu:
https://github.com/shaise/FreeCAD_SheetMetal/pull/407#issuecomment-2519408168

Help me to test to see it's OK

Seems strings were saved.

Also I marked some recently added strings and completed Spanish translation.


shaise:
Wow! This is quite some work. Thanks!
Looks OK to me.
I will probably remove FlangeAdvancedParameters.ui as it is not used anymore.
Are you going to update Crowdin with that?

hasecilu:
Already did :sunglasses:

shaise:
Thanks a lot!
I suspect we will have one more translation round by the time I finish all milestones.

Issue #426, opened 2024-12-21
Automatic unfolding of all parts in the model (SMUnfoldUnattended)


GS90:
### Describe the feature

Hi,
I'm not entirely sure that this is a "feature" and not a "bug", since we are talking about functionality that was already available...

Complex parametric models consist of dozens or even hundreds of sheet metal parts.

There is a need for convenient and quick translation of all parts into "blanks" (sheet metal flat patterns) for subsequent manufacturing.

Previously, the workbench had a command "SMUnfoldUnattended" that allowed you to use the unfolding function from a third-party script similar to mine, a simple example - https://github.com/GS90/addFC/issues/3.

Such functionality for processing sheet parts is quite rare in CAD systems, but very much needed in production tasks.

Now with the introduction of parametric unfolding, the "SMUnfoldUnattended" function has disappeared and this is a big blow to the functionality in general and especially to labor productivity...
Scripts for automatic processing of parts have stopped working.
Translating a hundred parts into *.dxf manually is a problem.

I have been working with sheet metal for quite a long time and sincerely believe that it is necessary to provide the ability to use the unfolding in automatic mode, and the user interface for the command and the "parametric unfold" itself are far from always necessary in work.

### Additional context

![fd](https://github.com/user-attachments/assets/b7b3c1bc-d9e5-47a1-8f0e-3f9eeaa5d779)

Here is an example of my model consisting of more than 50 different sheet metal parts, with the function of automatic export to *.dxf, the time for its processing by the method I described above is seconds.

### Would you like to work on this issue?

None

shaise:
hi @GS90 ,
Sorry about that, I can reinstate this functionality, can you share the part of your script that initiate this command? specifically, how do you select the faces/objects you want to unfold?

GS90:
A very simple algorithm:

1. Create a copy of the object in the active document, using a method similar to "Part SimpleCopy"
2. Search all faces and find the required one with the largest area
3. Select only the found face and call "SMUnfoldUnattendedCommandClass"

```
# reproduction:
shape = Part.getShape(body, '', needSubElement=False, refine=False)
ad.addObject('Part::Feature', 'Reproduction').Shape = shape
body = ad.ActiveObject

# find the largest face:
faces = body.Shape.Faces
target, a, n = [0.0, 0], 0, 0
for f in faces:
    a = round(f.Area, 2)
    n += 1
    if a >= target[0]:
        target[0] = a
        target[1] = n

# selection:
[...]

PR #430, opened 2024-12-30
Legacy compatibility (< 0.21) and object visibility (UnfoldUnattended)

GS90:
Hi @shaise.

A small addition for the ability to use older versions of FreeCAD. My previous change got lost :)

SMUnfoldUnattended: I suggest not hiding the highlighted object in this case #426.
I'm just giving him back visibility, maybe it's not very elegant...


shaise:
Good fix. Thanks!

PR #432, opened 2025-01-04
Removing export (UnfoldUnattended) and sketches recompute

GS90:
Hi,

**SMUnfoldUnattended** exported the drawing - this is clearly unnecessary.
I also added recalculation of created sketches so that they do not affect the document status.

### DIFF
--- a/SheetMetalUnfoldCmd.py
+++ b/SheetMetalUnfoldCmd.py
@@ -261,6 +261,7 @@ if SheetMetalTools.isGuiLoaded():
             for itemName in self.Object.UnfoldSketches:
                 item = self.Object.Document.getObject(itemName)
                 if item is not None:
+                    item.recompute(True)
                     objs.append(item)
             return objs
 
@@ -549,7 +550,6 @@ if SheetMetalTools.isGuiLoaded():
             SMUnfold(newObj, selobj, sel.SubElementNames)
             SMUnfoldViewProvider(newObj.ViewObject)
             SheetMetalTools.smAddNewObject(selobj, newObj, activeBody)
-            smUnfoldExportSketches(newObj, False)           
             selobj.Visibility = True
             return
 


shaise:
I thought that this was a needed function in export unattended? 
How do you know if others don't need it? this was the functionality before. If you do not want to export the sketches, uncheck the generate sketch.

GS90:
When I used the function before, it definitely did not save the dxf or svg file automatically.

If you uncheck the sketch generation box, they will not be created at all...
The question is specifically about exporting to an external format.
The sketches themselves are needed, but the need to export them to external formats is better left to the user's choice (manual export) or a special command, as has already been done in the **Unfold** graphical interface.

Currently, **SMUnfoldUnattended** does not have a choice of which of the sketches to save, for example, bend lines are needed in the drawing (TechDraw), but they are harmful when sending the workpiece for laser cutting. It's hard to imagine a use case for this.

In my batch processing, I export the required sketch myself, modifying it if necessary (for example, aligning it strictly in the middle).

shaise:
Well the previous functionality of unfold unattended was that if Export Sketch was active, the unattended was exporting it as well. As I did not like the fact that in standard unfold the export was done in the background without control of where to save the file, I changed it to manual export button.
However for unattended I wished to maintain compatibility. Can you say that there is no use at all for saving sketches automatically?

GS90:
I really don't see the point in automatically saving the sketch when using **SMUnfoldUnattended**.

Moreover, it is practically uncontrolled and implicit.
It was only after some time of use that I noticed it was saving something somewhere...

Examples of use, in my opinion:
* Checking whether the part can be unfolded and verifying the correctness of the resulting sketch.
* Creating a sketch for subsequent use in TechDraw.
* Using it for batch processing.

In all cases, a sketch is needed within FreeCAD, but its export is not required.
If necessary, the user can perform a controlled export (Ctrl+E).

shaise:
OK, then. I will accept the pull request, and if any user will have issues with it, I will add an extra checkbox

PR #437, opened 2025-01-11
Unfolder Command - Improve Pre-1.0 Check

Syres916:


shaise:
@Syres916 , do you think that having a replacement to `isParallel` can make the unfolder compatible with pre 1.0?
Also, as I understand whenever the addon manager install a new library dependency, it is installed in the user folder, meaning all freecad versions are sharing it. Does this cause conflicts if different python versions are used?
I'm asking since I want to understand why I can not import the new unfolder under 0.21 and you can.
Last thing, If we know for certain that we can not use it in pre 1.0 versions, wouldn't it be simpler just to check the freecad version?
Thanks for the fix!
shai

Syres916:
> do you think that having a replacement to `isParallel` can make the unfolder compatible with pre 1.0?

possibly but I need to do some compile testing with 0.21 to see if this is the only function that is being used

> Also, as I understand whenever the addon manager install a new library dependency, it is installed in the user folder, meaning all freecad versions are sharing it. Does this cause conflicts if different python versions are used?

Well it's only installing for that specific Python version and while _Help_ and some old _Plot_ addons are known to not work across different versions, the Python dependencies are usually stable in my experience. It's not going to be long before we're going to have Python 3.13.x users so getting comfortable with the variety is good preparation.

> I'm asking since I want to understand why I can not import the new unfolder under 0.21 and you can. Last thing, If we know for certain that we can not use it in pre 1.0 versions, wouldn't it be simpler just to check the freecad version? Thanks for the fix! shai

I'm getting less and less enthusiastic about version number checks, if we use attribute and method checks then it doesn't matter if Realthunder's branch is used or FreeCAD main etc we know whether the build is use has the ability to do the job but that's a personal opinion.

Syres916:
@shaise after a bit of back and forth using the 0.21.1 AppImage (Python 3.10 based) I installed `typing-extensions` using the AddonManager dependencies code as well as a new install of Sheetmetal Wb so `networkx` was installed as well. Then modifying the line `from typing import Self` to `from typing_extensions import Self` I got a clean import of the New Unfolder so the only thing left is to see if the `isParallel` can be given a Python alternative.

021_1_AppImage

Give me a day or two as I've some important paperwork to get done which will involve cleaning my desk, big job!!

shaise:
> if we use attribute and method checks then it doesn't matter if Realthunder's branch is used or FreeCAD main

Excellent point!! Thanks

alexneufeld:
> so the only thing left is to see if the `isParallel` can be given a Python alternative.

https://github.com/shaise/FreeCAD_SheetMetal/pull/438

shaise:
@Syres916 ,
It works now both on 0.21 and LS3.

PR #438, opened 2025-01-12
check vector parallelism manually for FC<=v0.21 compatibility

alexneufeld:


shaise:
Awesome! Thanks!

shaise:
Hope It's ok, but I changed the inline code into a function so later we can easily revert the the internal isParallel.

shaise:
@alexneufeld , is it possible to also remove the "from typing import Self"?
Edit, and also Vector.isNormal not available 21 and LS3

alexneufeld:
> @alexneufeld , is it possible to also remove the "from typing import Self"? Edit, and also Vector.isNormal not available 21 and LS3

:+1: :+1: https://github.com/shaise/FreeCAD_SheetMetal/pull/439

PR #439, opened 2025-01-12
polyfill isNormal, eliminate typing.self

alexneufeld:


shaise:
Thanks! Verified! working now on both 0.21 and LS3!

Issue #450, opened 2025-01-20
[NEW UNFOLDER] Issues with sketch and DXF on the new version.


thanhphong19512:
### Describe your issue

Hello, the latest new version works almost perfectly in many cases. However, there are still a few cases where I noticed issues. You can see them here:

**Case 1: Creating a sketch loses details (highlighted in the red box I marked).**

![Image](https://github.com/user-attachments/assets/9fa9ca30-0639-4048-8312-7407a98de7ee)

![Image](https://github.com/user-attachments/assets/878d7e2b-6157-4372-a421-1b7f57322fd6)

**Case 2: After exporting to DXF, if you re-import that DXF file into FreeCAD, you can see that it is not in the correct position.**

![Image](https://github.com/user-attachments/assets/459917e1-a588-40a5-890c-e2954737f027)

![Image](https://github.com/user-attachments/assets/1ce7bd3b-a02e-44a0-a775-4e401271e92e)

I’m attaching two files here for your reference.

[test.zip](https://github.com/user-attachments/files/18471701/test.zip)

### FreeCAD version info + SheetMetal WB version

```shell
OS: Windows 11 build 26100
Architecture: x86_64
Version: 1.0.0.39109 (Git) Conda
Build type: Release
Branch: (HEAD detached at 1.0.0)
Hash: 2fcc5317fe3aee96ca73475986a577719fc78e20
Python 3.11.10, Qt 5.15.15, Coin 4.0.3, Vtk 9.3.0, OCC 7.8.1
Locale: English/United States (en_US)
Stylesheet/Theme/QtStyle: unset/FreeCAD Classic/Qt default
Installed mods: 
  * sheetmetal 0.7.4
  * toSketch 1.0.1
```

shaise:
@alexneufeld , can you take a look?

thanhphong19512:
besides that the part that looks like the milling instead of sheet metal can still be unfolding

![Image](https://github.com/user-attachments/assets/f5ff0ff1-656a-494b-866b-2852b9b0f769)

![Image](https://github.com/user-attachments/assets/0151c623-948a-4eb6-8d80-ce1ec4d0c565)

[Part4.zip](https://github.com/user-attachments/files/18479782/Part4.zip)

shaise:
> besides that the part that looks like the milling instead of sheet metal can still be unfolding
> 

That's a nice feature!!

alexneufeld:
[...]

PR #451, opened 2025-01-21
Dxf export fix

alexneufeld:
partial fix for #450
- create sketches manually (instead of via Draft API) to ensure correct alignment
- don't skip first sketch layer if there are multiple to export


alexneufeld:
No match statements :sob:

shaise:
Our own sketch creator. Great!

shaise:
> No match statements 😭

Yes, I hate it. But turns out ver 0.21 install for win comes with python 3.8...

Issue #307, opened 2024-03-02
Making folded walls on planes or surfaces


pierreporte:
In sheet metal design, it is often needed to make flanges to follow certain planes or surfaces that were predefined, usually for enclosing or to support existing geometry. A workaround would be to draw walls at the precise dimensions so that `SMMakeWall` or `SMFoldWall` draw it perfectly where it should but it is not a practical workflow.

Other programs allow to use planes or surfaces as references. Usually, they need to cross the base wall. In other word, the base wall need to go further than the supporting geometry. The workbench would greatly benefit from this. It would be a mix between `SMFoldWall` and `SMMakeWall` because the workflow is similar to fold on line but it needs the parameters of the normal flange: leg length, chamfers, miters, etc. (except angle obviously).

![image](https://github.com/shaise/FreeCAD_SheetMetal/assets/2029853/ac4dc207-238a-47ab-96df-42ffb05a1cc1)

pierreporte:
@shaise What do you think of this feature?

It will need something to take into account zones that shouldn’t be bent as illustrated in [AddFoldWall](https://wiki.freecad.org/SheetMetal_AddFoldWall) wiki page. It would be a good idea that the UX is the same for both for these exclusions. I think that the best way would be to allow to select surfaces further than the computed fold line, and it would allow to solve #306 by simply allowing only one continuous line in the sketch.

shaise:
This can be a nice feature, I agree

pierreporte:
I think this feature could reuse part of `SMFoldWall` by computing a line automatically (but not creating a user-accessible sketch).

It could be a good idea to offer the ability of `SMMakeWall`, that is asking for a leg length, and the ability of `SMFoldWall` that just folds existing geometry. There could be a third option using a sketch that would become the contour of the new flange.

@sheetmetalman if you’re looking for something new, this is one of the last big missing features of the workbench.

sheetmetalman:
@pierreporte Hi!

Maybe I got the idea. Maybe not. Tell me if I got the idea right, please.

Maybe it can be done with the `SMMakeWall`. Just need to be added another type of parameter for offset and angle. Instead of only values, can has something more advanced like face selection.

For the angle: the user select a face as a parameter and select if the angle must be parallel or normal to that face.

For the offset: the user select the same face, and the limit will be the intersection of the sheetmetal plane with the plane of the user selected face, on other words the distance between the end of sheet metal to intersecttion.

[...]

PR #452, opened 2025-01-24
Improve bend offset calculations when angle isn't 90deg.

karliss:
TODO: will prepare test project and screenshots tomorrow.

Previously when doing bend using "Make wall" command "Material Inside" and "Material Outside" modes didn't take into account the angle.

Test plan:

- Using Outer sharp or Inner sharp modes
    * Material inside, 90 > angle > 0
    * Material inside, angle > 90
    * Material inside, -90 < angle < 0
    * Material inside, angle < -90
    * Material outside, 90 > angle > 0
    * Material outside, angle > 90
    * Material outside, -90 < angle < 0
    * Material outside, angle < -90
    
 In each of the cases above changing radius should not affect endpoint
  
 - Using inner sharp mode change angle from -145 to 145
 - Using outer  sharp mode change angle from -145 to 145
 - Using tangential mode change angle from -145 to 145
 
 During the 3 cases described above the shape should change smoothly, negative angles should look the same as positive angle reversed.

### DIFF
--- a/Resources/panels/FlangeParameters.ui
+++ b/Resources/panels/FlangeParameters.ui
@@ -1408,6 +1408,26 @@ min-width: 28px;</string>
                    </iconset>
                   </property>
                  </item>
+                 <item>
+                  <property name="text">
+                   <string>Material Inside (90° legacy)</string>
+                  </property>
+                  <property name="icon">
+                   <iconset>
+                    <normalon>Icons:SheetMetal_WallPosMatIns.svg</normalon>
+                   </iconset>
+                  </property>
+                 </item>
+                 <item>
+                  <property name="text">
+                   <string>Thickness Outside (legacy)</string>
+                  </property>
+                  <property name="icon">
+                   <iconset>
+                    <normalon>Icons:SheetMetal_WallPosThkOut.svg</normalon>
+                   </iconset>
+                  </property>
+                 </item>
                 </widget>
                </item>
               </layout>
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -1263,10 +1263,16 @@ def smBend(
     if BendType == "Material Outside":
         offset = 0.0
         inside = False
-    elif BendType == "Material Inside":
+    elif BendType == "Material Inside": # legacy mode for backwards compatibility
+        offset = -(thk + bendR)
+        inside = True
+    elif BendType == "Thickness Outside": # legacy mode for backwards compatibility
+        offset = -bendR
+        inside = True
+    elif BendType == "Material Inside V2":
         offset = -(thk + bendR) * abs(math.tan(math.radians(bendA * 0.5)))
         inside = True
-    elif BendType == "Thickness Outside":
+    elif BendType == "Thickness Outside V2":
         offset = -bendR * abs(math.tan(math.radians(bendA * 0.5)))
         inside = True
     elif BendType == "Offset":
@@ -1781,7 +1787,9 @@ class SMBendWall:
             obj,
             "BendType",
             translate("App::Property", "Bend Type"),
-            ["Material Outside", "Material Inside", "Thickness Outside", "Offset"],
+            ["Material Outside", "Material Inside V2", "Thickness Outside V2", "Offset",
+             # legacy modes
+             "Material Inside", "Thickness Outside"],
         )
         SheetMetalTools.smAddEnumProperty(
             obj,
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -1264,10 +1264,10 @@ def smBend(
         offset = 0.0
         inside = False
     elif BendType == "Material Inside":
-        offset = -(thk + bendR)
+        offset = -(thk + bendR) * abs(math.tan(math.radians(bendA * 0.5)))
         inside = True
     elif BendType == "Thickness Outside":
-        offset = -bendR
+        offset = -bendR * abs(math.tan(math.radians(bendA * 0.5)))
         inside = True
     elif BendType == "Offset":
         if offset < 0.0:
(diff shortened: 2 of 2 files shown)


shaise:
Thanks!
Remarks:
1. Can you attach an image of what do think should be the new dimensions if the angle is not 90 degrees?
2. We need to take into account that changing the results of the measurements might break old files that are counting on the current behavior. So we might need to add new modes instead of replacing the current ones. 
3. Please stand by with this PR until the bend wall new features are implemented. See Issue #307

karliss:
I feel like might need a few more testcases to convince myself of the new behavior being more useful. Here is what I did so far. 
Let's start with the "good" case.

The shape is created by making flat 30mm wall, 140 degree corner 20mm wall, 40 degree corner with 30mm length, -140 degree corner selecting outside with 20mm lenght. First row is using  "Material Inside"+"Outer sharp", second row is "Thickness outside"+"Inside sharp". Of  course in practice this specific part would be easier to create by using base bend feature, but the same problem would apply for more complex 3d shapes which you couldn't create using base bend from sketch. Only difference between left column and right column is radius.
Note how before patch, the shape changes based on radius, which should not unless "Material Outside" or "Leg"  modes are used. After changes it's easy to get desired dimensions and changing radius only affects the radius and nothing else. Also before patch, length for negative angle is completely wrong.
**Upstream**
p2_upstream_705
**Patched:**
p2_modified

Now here comes the bad case.

First row "Material inside", second row "Thickness outside". 1st column base radius 1mm side radius 1mm, 2nd column base radius 1mm, side radius 2mm, 3rd column base radius 2mm side radius 2mm,  4th column base radius 1mm side radius mix of 1mm and 2mm. 

In this case before patching when all radius are identical due to shape of parallelogram errors partially cancel out so that the two sides at least align, positioning is still off and varies based on radius, negative angle length completely wrong.
In the patched version changing radius doesn't affect position of walls, but they don't align due offset introduced when creating walls from the base,  

I have a feeling that if the base shape was different polygon the seeming advantage of upstream behavior might disappear.

**Upstream**
image
**Patched:**
parrallelograms_modified
**Source of misalignment in patched version**
image




What to do next -> compare the behavior in similar setups with all 90deg angles. More complex polygon.  Come up with other cases and test how the new and old behavior makes sense.  I am slightly worried that the offset which happens in base bend, might significantly reduce usefulness of the changes I am proposing.

karliss:
One more observation reopening the project with modified sheetmetal WB and running recompute doesn't cause the sheetmetal part to change. Only if I modify some property of base shape like the material thickness it actually recomputes and updates the shape. Saying this only for the purpose of testing, I am not suggesting that relying on this for project compatibility.

luzpaz:
This looks like a complex effort. Thank you for all the work you've done so far :pray:

karliss:
Now that face reference changes are done I guess this is unblocked. Next steps would be:
* rebase on top latest changes and look into whether any of wall changes affect this PR
* project backwards compatibility

Like suggested i am thinking of adding two new modes and renaming the old modes since any new project should probably prefer new modes. I hope that the mode names in project and UI strings are separate. If the list of modes gets's too long/messy then the legacy modes can be hidden by default showing only for objects which had chosen them in older workbench versions.

shaise:
Yes. Hope rebasing will not make many conflicts

karliss:
Rebased and added the fixed calculations as new modes. Within task GUI old modes are labelled as "Thickness Outside (legacy)". Unfortunately the property view directly displays the internal names used by code, so it's a bit more confusing there. But the property view also doesn't use localized strings.

After playing with the feature more noticed that there is one aspect of non 90° that would be good to discuss early, as attempting to fix later will introduce more breaking changes in behavior.

If you compare to face reference modes, setting angle to something like 45° essentially establishes a plane. The non obvious part is which of the two planes (going through left corner, or right corner of wall before adding the new bend).

I can think of at least 3 if not more somewhat logical ways it  could behave:
a) assume previous wall is on same side -> if you choose material inside assume that previous wall was already inside
b) define plane based on the corner user picked before running bend command
c) top hit -> imagine a plane moving towards unbent wall, which ever corner was hit first stop there. Might be a bit non-intuitive when applied to thickness outside. Maybe "top hit" for material inside and "bottom hit" for thickness outside.

For 90° bends there is no choice since there is single plane going through both corners.

For a) and c)  has nice property that it matters less how exactly bend was created. 
b) is slightly more powerfull as it allows user to choose which of the 2 planes the wall should be inside or outside of. 

image
image

Maybe I should try implementing b) and c) to see how useful they are in practice when you try to more complex shapes.

shaise:
IMO the most important thing is that it is consistent and simple to explain. I does not have to be powerful or make assumptions, or relay on user selecting edges.

karliss:
> I does not have to be powerful or make assumptions

You can't hand wave away inherent choice. There are 2 corners and math needs to pick one. Either the code makes the choice for user, picking the one which we assume will be more useful in most situatuions. Or the user needs to pick themselves implicitly (like the edge which gets selected to start the bend) or explicitly with additional UI element.

And if the choice is done badly, the tool becomes useless. 

After a bit of thinking I am somewhat sure that the approach c) doesn't work. Depending on angle it will pick the wrong corner.

For the approach a) how about tweaking icons to something like this:
image


As for the possibility of explicit choice that could be merged with the offset reference face activation, thus avoiding introduction of more dropdowns than we already have.

Currently there are 2 dropdowns containing nearly identical choices:
* Wall pos 
   - Material Outside
   - Material Inside
   - Thickness outside
   - Offset
* Offset position (shown only when offset face is selected)
   - Material Outside
   - Material inside
   - Thickness outside
   - Offset

The alternative d) could be having :
* Bend reference
   - Inside corner 
   - Outside corner
   -  Face (similar effect as currently selecting offset+reference face)
* Wall pos
   - Material Outside
   - Material Inside
   - Thickness outside
   - Offset

karliss:
Currently I am leaning more towards a).

Seems like each CAD uses slightly different naming and each provides different subset of choices:

| Freecad naming+       | Fusion 360 | Solidworks              | Onshape | 
|---------------------------------------------------|------------|-------------------------|-------------------------|
| Material Outside                                  | Adjacent   | Bend Outside            | Hold line |
| Material Inside (old mode)                        | Tangent    | Tangent to bend         | - |
| Thickness Outside (old mode) | - | - |  - |
| [new version] Thickness outside (Inside corner)              | Outside    | Material Outside        |(align) Inside [???]
| [not implemented] Thickness Outside (Outside corner)  | -          | -         |   |
| [not implemented] Material inside (Inside corner)  | -          | Material Inside         | 
| [new version] Material Inside (Outside corner)               | Inside     | Bend from Virtual Sharp | (align) Outside [???]
| - | - | - | Middle  |

??? - it's unclear whether Onshape Outside behaves like "Material Inside (Outside corner)", "Material Inside (Inside corner)" or something else since the docs only showed 90° flange, same for "Inside". 

Autodesk inventor -> 4 choices corresponding to main modes +  hidden checkbox called "Old method" which based on illustration probably changes the inside corner/outside corner choice.

I had noticed before that the old "Material Inside" behavior for sharp angles made at least some sense.  Should probably rename it from "Material Inside (legecy)" to something else, might be worth keeping it not just for backwards compatibility.

luzpaz:
Hey @karliss what is the status here ? Are you still working on this ?

shaise:
Sorry, did not work on it

karliss:
I am not actively working on this, since I currently don't have a use for sheetmetal.

Feel free to take over the work on this if you are interested. I am still available to discussions or questions if you have any.

PR #455, opened 2025-01-27
Improve unfold solid generation

### LINKED ISSUE #450: [NEW UNFOLDER] Issues with sketch and DXF on the new version.

alexneufeld:
add additional null-checks and cleanup passes, as well as constructing
the final solid from a single face, rather than using boolean operations.

closes #450

Extra details:

FreeCAD's (OpenCascade's?) tolerance requirements for shapes to pass a BOP check are that errors (misalignment between points, gaps in surface seams) are less than 0.1 NANOmeters. In my experience with test cases submitted by users in recent issues, as well as my own models, it's common for FreeCAD to produce geometry with errors on the order of 10-20 nm, which is much higher than the required tolerance. Using a Part::3DOffset or the sheetmetal sketched cutout tools are both good ways to create 'bad' geometry. unfortunately, the FC Gui isn't very good at communicating to the user when they have done something to create out-of-tolerance geometry, and there aren't many tools to repair bad shapes anyway. We can probably continue to expect to see people complain that their parts won't unfold, not realizing that the part geometry is garbage as far as OCC is concerned. We could run a BOP check every time we unfold a shape, but that's a fairly slow operation (which is why FreeCAD does not do it automatically in most cases). For now, the additional work in this PR actually makes some invalid shapes unfold, provided the user selects a favorable root face. :+1:

### DIFF
--- a/SheetMetalNewUnfolder.py
+++ b/SheetMetalNewUnfolder.py
@@ -44,10 +44,28 @@ except ImportError:
         "or reinstalling the SheetMetal workbench using the addon manager\n"
     )
 
-# used when comparing positions in 3D space
+
+# we need to VERY CAREFULLY choose multiple different 'epsilon' values for
+# different types of numerical comparisons
+#
+# default eps value to use for most numerical approximations.
+# This is intentionally larger than OCC's tolerance requirements,
+# so that out of tolerance geometry can still be processed
+# and then fixed later with cleanup passes
 eps = FreeCAD.Base.Precision.approximation()
-# used when comparing angles
+# this is used instead of 'eps' when comparing angles
 eps_angular = FreeCAD.Base.Precision.angular()
+# when running cleanup passes, it will be assumed that points that are
+# closer together than this value should be altered to be exactly coincident
+fuzz = 1e-3  # <-- 1 / 1000 * 1mm  = one micrometer
+# this is OCC's upper bound for tolerance errors when building geometry.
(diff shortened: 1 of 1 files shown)


shaise:
Yes! anything that saves boolean operations is a blessing!
Thank you!

PR #457, opened 2025-01-31
fix issue with bend line placement

alexneufeld:
when not using seperate sketch layers, the bend lines would be placed incorrectly


shaise:
Thank you!

PR #471, opened 2025-02-11
Fix selection filter in AddWall command

sheetmetalman:
User was able to select a face as reference to create a wall before the click on AddWall command.
With this fix, the user is no longer able to do this.

In old files where the user used a face as a reference to create the Wall, the tool will work normally. Because the code was kept to be compatible with old files.

Also, fixed the tip of AddWall command button.

[Related to this forum topic.](https://forum.freecad.org/viewtopic.php?p=810310#p810310)


shaise:
Thanks!
I was thinking for some time to remove the face selection option. I think it is a good idea since its more predictable.
The wiki already is updated to use edges and not faces.
The only issue is what happens if you want to fold a flange up to a face that is inside the same solid as the fold. In this case if you select the face in advance you can not perform this operation.
This is not a big issue, since you can always do it through the task panel later.

sheetmetalman:
> The only issue is what happens if you want to fold a flange up to a face that is inside the same solid as the fold. In this case if you select the face in advance you can not perform this operation.
> This is not a big issue, since you can always do it through the task panel later.

I agree! There is this problem! I saw this when fixing this, but I came to the same conclusion. :sweat_smile:

jaisekjames:
By using clear selection & press preview gives error [due to no edge in selection].
We need to change the gif in github loading page as it shows face selection.

jaisekjames:
We can use selection Filter gate so that only edges only selected.
https://wiki.freecad.org/Selection_methods

sheetmetalman:
This happen with old files. Because the new code only accept edges as a selection. When you empty the selection and click "preview" the code appear try to add again the old reference, but the code doesn't accept that. Then error!

Currently, before click in AddWall is impossible to select an edge. And in new files is impossible to select an edge throug taskpanel (what I see in this case is that the user can select a face but when click in the button preview that face isn't added to references).

I think, what we need is a code to deal with old files on face selection through taskpanel. Maybe something like this: if the old reference is a face, when user click "preview" the code simple add an edge that belongs to that face, but this edge cannot be the thickness edge.

PR #474, opened 2025-02-22
SM: Change AutoColor of created sketches to False

Roy-043:
Without this colors of sketches are automtically changed to match the current preferences.


shaise:
Thanks!!

PR #475, opened 2025-02-22
Correct AutoColor 0.21.x & LinkStable compatability

Syres916:
Also added same to the new unfolder, CC @Roy-043


shaise:
Thanks again!

Issue #469, opened 2025-02-05
Create Unfolding as new body


jaisekjames:
### Describe the feature

If we can generate auto unfolding in new Body with shape-binder, shape will  unfold automatically & we can keep adding new features.

![Image](https://github.com/user-attachments/assets/41d810e8-6b51-43b2-8e4b-24bb67a0cd2b)

**Note:** Sample file attached. Rename file [remove .zip]

[Auto unfolding.FCStd.zip](https://github.com/user-attachments/files/18669918/Auto.unfolding.FCStd.zip)

### Would you like to work on this issue?

None

jaisekjames:
Note: Change Set tip/Suppress features to see unfolding changes.

shaise:
You are correct! and this is on my list. I thought of making it as a simple object, not a body, though.
Did you make it automatically or manually?
Also, do you think there is a good reason for it to be under body and not just a simple separate object?

jaisekjames:
I made it manually. When I saw Autodesk Inventor screenshot, they are using same tree structure [ like two separate body. This one rename to flat pattern.]

![Image](https://github.com/user-attachments/assets/0dd3a04c-9faf-4886-87b5-acb7c8ea2842)

FC-FBXL5:
What about an automatic switch? 
If the object to unfold is inside a body, the Unfold solid and maybe the sketches are put into a new body for a PartDesign workflow, or else independent objects are created for a Part workflow.

shaise:
@FC-FBXL5 this was my initial thought, but putting it in another body with all the extras (such as the binder, origin and extra tree level), just for a single solid does seem to be a logical outcome.

Mannshoch:
In Solidworks the Unfold is a separate Child configuration (Update value with the main body) which is usually used by Drawings. In model is not possible to have both bodies at the same time, but in a Drawing you could show different configuration per view.
Therefor you have the possibility to create several Unfoldings for the same body. e.g. 

- If you wish to show the folding process in several steps you could duplicate the configuration for as many steps you need.
(therefor every Unfold Feature in tree has per bow a child feature which could be deactivated) 
- You could visualize production steps like an unfolded and an unfolded and punched case

pierreporte:
I agree with @Mannshoch. The unfolded configuration should be some sort of state, not a new body.

shaise:
[...]

PR #486, opened 2025-03-22
Separation of Name and Label for Unfold

GS90:
Hi,
Fix for replacing non-Latin characters in the created Unfold object.
https://github.com/shaise/FreeCAD_SheetMetal/issues/469#issuecomment-2740822016

image


shaise:
Nice fix! thanks!

PR #487, opened 2025-03-24
If user has an old version of NetworkX without Graph attribute use le…

Syres916:
…gacy Unfolder

Based on feedback from forum https://forum.freecad.org/viewtopic.php?p=817856#p817856 give user feedback to update NetworkX version.


shaise:
Thanks!!

PR #497, opened 2025-07-01
Add icons to top titles of TaskPanels

zokhasan:
Icons_TaskPanels

### DIFF
--- a/ExtrudedCutout.py
+++ b/ExtrudedCutout.py
@@ -412,6 +412,7 @@ class ExtrudedCutout:
 ##########################################################################################################
 
 if SheetMetalTools.isGuiLoaded():
+    from PySide import QtGui
     from FreeCAD import Gui
 
     icons_path = SheetMetalTools.icons_path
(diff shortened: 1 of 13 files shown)


shaise:
I like this idea!!

However if it is OK with you, I would prefer to embed the icons directly in the UI file, without using code. Here is an example:

image

Here is the actual text:
```
  <property name="windowIcon">
   <iconset>
    <normaloff>Icons:SheetMetal_AddBaseShape.svg</normaloff>Icons:SheetMetal_AddBaseShape.svg</iconset>
  </property>
```

It it OK with you?

zokhasan:
It's interesting idea. I don't know how better. At this step I assume that it will be better to add in the code, because the code already has the same links (in neighboring classes) and perhaps they'll be combined into one link in the future. If this is added to UI then the same link will be in two different files: in py and ui.

shaise:
hi @zokhasan ,

You are right that these Icons are already used in the code. This gave me an idea to set the icon automatically, this way there is no repeat of the icon name and only 3 file changed.

See commit: https://github.com/shaise/FreeCAD_SheetMetal/commit/1fa547a19d7626a441e87088a12acb1e8147e28d

I have credited this change to you in the README .
Thank you!

PR #498, opened 2025-07-04
Fix assign color for lines in UnfoldTaskPanel

zokhasan:
Fix assign color for Internal and Fold lines. A bug was caused that a color selected in UnfoldTaskPanel for Bend lines was assigned for Internal lines and vice versa.

fix_color_lines


shaise:
Nice catch! Thanks!

PR #500, opened 2025-07-06
Refactoring ExtrudedCutout.py

zokhasan:
Made the using of some if-statements more optimally, removed redundant checks for assigning the same variables.


shaise:
Thanks! I always like it when code can be aggregated.

PR #501, opened 2025-07-08
Remove duplicate class SMLogger

zokhasan:
Same SMLogger classes were in SheetMetalLogger.py
and SheetMetalTools.py. The newer commit 3a793093 (8 Dec 2024) copied this class in SheetMetalTools.py, so class is remove from SheetMetalLogger.py.

Correct imports of this class in other files.


zokhasan:
In
- SheetMetalFoldCmd.py
- SheetMetalFormingCmd.py
nothing used this class, so imports just deleted.

Lines in SheetMetalBaseShapeCMD.py that use this class were commented,
so make import also commented.

shaise:
Thanks for the cleanup!
I have removed SheetMetalLogger altogether.

zokhasan:
@shaise , 'smwb_locator.py' has also only a license notice

shaise:
Indeed.
@easyw , do you remember why you added this file?

Issue #481, opened 2025-03-08
The workbench lacks a sketch creation command


pierreporte:
### Describe the feature

The sheet metal workbench works with sketches, but it doesn’t offer a command to create them. Users need to either go to PartDesign or Sketcher workbenches to do so. It would be a good idea to provide one of the two commands, `Sketcher_NewSketch` or `PartDesign_NewSketch` (preferably the PD version). Once https://github.com/FreeCAD/FreeCAD/issues/12106 is resolved, the new unique command could replace it.

### Would you like to work on this issue?

None

shaise:
Agree completely. Added to my list.

zokhasan:
https://github.com/user-attachments/assets/acd18903-9f9f-4d99-92a1-4e21a6c52f28

pierreporte:
@zokhasan of course it can be added manually but it should be there by default.

shaise:
anyone knows how to do it by code?

davesrocketshop:
Import the Sketcher commands

`import SketcherGui`

Add it to your toolbar (example from the Rocket WB)

`self.appendToolbar(translate('Rocket', 'Rocket'),
                        ['Separator', 'Rocket_NewSketch', 'Sketcher_EditSketch', 'Separator', 'Rocket_Calculators'])`

davesrocketshop:
I also have a command to create a new sketch in the XZ plane (Rocket_NewSketch in the previous post). You can see the code for that here:

https://github.com/davesrocketshop/Rocket/blob/master/Ui/Commands/CmdSketcher.py#L46

zokhasan:
I was able to do the same as in the video (like "Customize…" command) but through code.
What other buttons should be on this extra panel by default? Maybe "Create body",... or only "Create sketch"?

shaise:
The problem is the sketch button should be smart to know if to start a partdesign sketch or a part sketch

zokhasan:
@shaise , as I understand it, if add button "Create sketch" from the Sketcher workbench (not from PartDesign workbench) then it will work with both PartDesign and Part... or I didn't catch up with something...

zokhasan:
I mean, for SheetMetal workbench it might be enough to add only "create sketch" from Sketcher workbench..

shaise:
I'm not sure. I dont know how it will work of the sheetmetal is in a body.

zokhasan:
[...]

PR #503, opened 2025-07-12
Add 'Create sketch' command (issue #481)

zokhasan:


pierreporte:
I don’t think it’s a good idea to add the command only if it’s already present.

@shaise what do you think?

shaise:
@pierreporte , please explain. what do you mean if it's already present?

pierreporte:
@shaise The code checks if `Sketcher_NewSketch` or `PartDesign_NewSketch` are already present on the toolbars, and adds `PartDesign_NewSketch` only when they are not already there.

shaise:
@pierreporte , Yes, and what do you think should be the correct behavior?
In any case, I'm against setting it as a separate toolbar, it should be added as part of the main sheetmetal toolbar. Also the best thing is to automatically detect if it is a part sketch or partdesign sketch. if not, this should at least somehow be clear to the user what is he using. The icons are too similar

pierreporte:
I would prefer that it’s added regardless of what the user already added to the toolbar, because it’s a UI/UX design choice, so it seems we agree.

As for the two similar commands, the best solution is to merge the two commands so that `Sketcher_NewSketch` can do everything: https://github.com/FreeCAD/FreeCAD/issues/12106

luzpaz:
@shaise so to summarize you agree with the idea but not with the application of it ?

shaise:
> @shaise so to summarize you agree with the idea but not with the application of it ?

yes

luzpaz:
> In any case, I'm against setting it as a separate toolbar, it should be added as part of the main sheetmetal toolbar. Also the best thing is to automatically detect if it is a part sketch or partdesign sketch. if not, this should at least somehow be clear to the user what is he using. The icons are too similar

@zokhasan what are your thoughts. Would you consider modifying it. Also considering the unknown future of https://github.com/FreeCAD/FreeCAD/issues/12106

zokhasan:
Sorry @luzpaz , I just saw this message.
Honestly I'm not sure about this path, so I'll pass.

velog:
It is must feature to have sketch on the tool bar, and maybe some other features from the Part Design(fillet, chamfer and patterns)

pierreporte:
@shaise This PR should be closed since it was superseded by commit 5b18132e6e5ebf3990b3b509bc40eeac88c33ba9

PR #504, opened 2025-07-13
Fix enable params when open the BaseShapeTaskPanel

zokhasan:
https://github.com/user-attachments/assets/3e38c374-20f8-41a0-a09d-e697776f1fd5


shaise:
Thanks!!!!

PR #505, opened 2025-07-14
SheetmetalTools - keep our Windows LibPack 0.21.x users working

Syres916:
@shaise this came from a report on the forum from a Windows 0.21.2 LibPack build user which still has Python 3.8.x see https://forum.freecad.org/viewtopic.php?t=98297 hope this is OK.


shaise:
If this is the only change, its worth it. Thanks!

PR #506, opened 2025-07-15
Change BaseShape: hide setting instead disable(Closes #355 issue)

### LINKED ISSUE #355: Feature: Hide Base Shape Type setting (instead of disabling)
Changing Base Shape type could hide the entire line instead of just disabling the input field. 
I think there is a property for that. 

What do you think? This would improve the usability a bit. 

image

zokhasan:
Also rename updateEnableState() on updateWidgetsVisibility() like in 'ExtrudedCutout.py'.

https://github.com/user-attachments/assets/60a0c024-94d4-4c72-9581-6bb9163be1cc

### DIFF
--- a/Resources/panels/BaseShapeOptions.ui
+++ b/Resources/panels/BaseShapeOptions.ui
@@ -91,376 +91,460 @@
     </layout>
    </item>
    <item>
-    <layout class="QHBoxLayout" name="horizontalLayout_5">
-     <item>
-      <widget class="QLabel" name="label_2">
-       <property name="text">
-        <string>Thickness</string>
-       </property>
-      </widget>
-     </item>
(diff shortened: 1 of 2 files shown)


shaise:
Sorry for the late reply.

What are the changes in the .ui file? the diff is too big to understand.

shai

zokhasan:
I placed the param lines in widgets and set zero indentation (Margin) in the settings of these widgets. I didn't correct code ui file manually, so indentations in other lines this ui file were automaticallu corrected. Line `<enum>Qt::Horizontal</enum>` was replaced by `<enum>Qt::Orientation::Horizontal</enum>` also automatically.

shaise:
looks good! thanks!

PR #510, opened 2025-08-01
Replace OrigButtons with RadioButtons(BaseShape)

zokhasan:
- In BaseShapeTaskPanel replace OrigButtons with RadioButton. The new widget is drawn in UI-file;
- Aligned SpinBoxes of param to widget with radio buttons.

The OrigButtons has display problems: red selection frame isn't set outside the red circle(coordinate position designation), stretched red circle in oval, gaps between buttons pictures.

-------

https://github.com/user-attachments/assets/ac24dcff-ac46-48d0-a2eb-87e9c1ce5a77

-------

Replace orig buttons with radio buttons (BaseShape)

The current option has display problem, sach as the red selection frame isn't set outside the red circle, but is set in the center of the button.  

01_current

This pull request offers to replace the Buttons with Radio Buttons, which will solve this problem. The background is drawn in a UI file by three QFrame widgets placed overlaid on top of each other: with dark outline color at the bottom, with highlight color in the middle, and main color on the top. What it looks like in Linux Fedora42 ...  

02_color_fill_linux

... and what it looks like in Windows10

03_color_fill_windows

-------

**P.S.**: First I tried a gradient background color, but then the radio button background becomes visible on the main background...  

04_rbbackground  

... and if turn off radio button background, then in FreeCAD Classic Theme (in Linux) the radio buttons will be invisible

05_rb_withoutbackgroungrb


pierreporte:
Your before situation looks quite strange, but in any case the new UI looks better than the current one theoretically does.

shaise:
I personally do not like the radio button look, specially in the classic theme. I would rather fix the scaling issue with the original widget. 
However, if the DWG decides this is better I will accept it.

PR #517, opened 2025-08-17
Refactoring SheetMetalTools.py

zokhasan:


shaise:
Thank you.
I see that sometimes you add spaces to `=` texts:
image
and sometimes you remove:
image

is this the common way?

zokhasan:
Thank you for asking) 

PEP8 suggests in common way to surround binary operators with a single space on either side, but at the same time PEP8 says:

> Don’t use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter:
> 
> ```python
> # Correct:
> def complex(real, imag=0.0):
>     return magic(r=real, i=imag)
> ```
> 
> ```python
> # Wrong:
> def complex(real, imag = 0.0):
>     return magic(r = real, i = imag)
> ```

shaise:
Great! Good to know!

PR #520, opened 2025-08-31
Fix various typos

luzpaz:
Not sure why vscode auto-formatted `SheetMetalFormingCmd.py`
I only fixed `./SheetMetalFormingCmd.py:100: faild ==> failed`

@zokhasan Any idea what changed in the refactor (ef2d459) that would cause this ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ SheetMetal WB can be installed via the [Addon Manager](https://github.com/FreeCA
                       - Update translations by [@hasecilu][hasecilu].  
                       - Fix python compatibility by [@Syres916][Syres916].  
                       - Extruded cutout fixes by [@sheetmetalman][sheetmetalman].  
-                      - Fix round relief apect ratio limitations.  
+                      - Fix round relief aspect ratio limitations.  
 * V0.6.00 03 Dec 2024:  Code refactoring, Remove duplication, Add GUI to most functions.  
 * V0.5.10 25 Nov 2024:  Extruded Cutout: add improvements and fixes by [@sheetmetalman][sheetmetalman].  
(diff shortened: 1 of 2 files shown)


shaise:
@luzpaz , I see a lot of non real diffs, probably whitespaces, but I don't know what kind.

zokhasan:
I know what the problem is, but it's not because of my changes.

As I understand it, the problem is that the repository have different line separators in the file('\n' and '\r\n'). For example, I added only two lines in SheetMetalFormingCMD.py  and the rest change on their own due to the changes in the end line separator.
example

It's not just this file, some others project files have the same problem. And when GIT gets this file, it doesn’t understand what to do with it.

When you look at the changes with `git diff` before `git add` it shows these lines with a "^M" mark. If you have a OS Windows and you see these marks, then you can install Notepad++ and manually change "\r\n" to "\r" in these lines. It's tiring, but I use this way so that the changes wouldn't show what I didn't change. It would be better to just update all the end separators with a separate commit, make them the same. And maybe this will solve the problem until anyone input another separator again.

luzpaz:
Thanks for the explanation @zokhasan  
I just went ahead and reverted ./SheetMetalFormingCmd.py:100: faild ==> failed  
@shaise this PR is ready

shaise:
Done. Thanks!

luzpaz:
@shaise looks like it's important to decide what the official take should be on line separators, as mentioned above by @zokhasan. Because the next edit from someone is going to most likely have the same response that I got.

shaise:
Yes, I agree. different programmers use different text styling, and I believe there is no styling that is objectively better than another.

PR #531, opened 2025-09-18
Remove "Add translations path" sections (except InitGui.py)

zokhasan:
Removing "Add translations path" sections from all SM WB files except InitGui.py, since only one is enough.


shaise:
Oh, Thanks! i did not know that.

Issue #412, opened 2024-12-04
SheetMetal v1.0 Milestones


shaise:
### Milestones for ver 1.0

- [x] Remove code duplication. Add helper function for common tasks
- [x] Add GUI (task panel) to all functions
- [x] Sticky parameter values when a function is reused
- [ ] Add and edit tool tips
- [ ] Make moving the tip and inserting of sheetmetal features in PD work 
- [x] Parametrize unfolding (using manual update button, because it is slow)
- [ ] Update translation files on crowdin
- [ ] Make the WB TNP - aware? https://github.com/FreeCAD/FreeCAD/issues/18671
- [x] AddBase: keep sketch visible #421 
- [x] Properly propagate object's material. 
- [x] Make forming tool easier to use + add missing task panel
- [x] Add cutout missing task panel
- [x] Solid body/Shell to SheetMetal #385 
- [x] New unfolder by @alexneufeld !
- [x] Option to detach the unfolded object from the body
- [x] Improve flange command by @sheetmetalman 
- [x] Add option to delete edges from multi selections in the task panel
- [x] Add sketch command to the toolbar. Issue #481 
- [ ] Allow commands to be used without prior selection. Issue #480 
- [x] Split functionality of `SheetMetal_Extrude`  command. Issue #482 
- [ ] Add intersection clearance for standard extend (similar to extend by sketch). Issue #560  

### Other milestones:
- [ ] Add preview geometry. [Based on freecad PR](https://github.com/FreeCAD/FreeCAD/pull/29641)

shaise:
@hasecilu,
I have implemented GUI panels to all missing function, Can you please update Crowdin with these changes?

hasecilu:
Hi, file has been updated.

ma00aem:
How can i test this release? Is it updated automatically or how to download it?

shaise:
Everything marked "V" are already in the latest release.

luzpaz:
@shaise any plan to roll this out around the time of the  the FreeCAD 1.1 release ?

shaise:
out of all the issues, the TNP is the most important. It looks like someone is working on it.
Regarding other milestones, I can try to attend some, but even if I dont, if TNP is fixed, I think we can mark it as V1.0

luzpaz:
> Regarding other milestones, I can try to attend some, but even if I dont, if TNP is fixed, I think we can mark it as V1.0

Makes absolute sense :+1:

[...]

PR #534, opened 2025-09-22
Improve TaskPanels: Sync with select objs in 'Select' mode

zokhasan:
Improve TaskPanels: Sync with selected objs in "SELECT" mode.

https://github.com/user-attachments/assets/b8274cf4-eb98-443f-bac0-c1043260e686

The change is already able to work, but I'm still going to check
if it's possible to remove other duplicating observer func and
classes in SheetMetalTools.py

I also would welcome to see feedback on how I connected this class in task panels.
In TaskPanels:
```python
class SomeTaskPanel:
    def __init__(self, obj):
        #...
        self.form.AddRemove.clicked.connect(self.select_clicked)

    def select_clicked(self):
        SheetMetalTools.SelectionObserver.manager(task_panel=self)
```
In the panel class when user click the `Select` button I call the `select_clicked` method in wich i send the `self` value as attribute.This is somewhat similar to the delegation approach, but with the self attribute being sent.
I'm comfused about sent "self" for an other class. But in the same time I like the ability to connect in 3 lines to any panel. What do you think about it?

-----------

In [SheetMetal v1.0 Milestones #412](https://github.com/shaise/FreeCAD_SheetMetal/issues/412) has the item "`Add option to delete edges from multi selections in the task panel`". How do you imagine it: like this or like this or how else?

### DIFF
--- a/SheetMetalTools.py
+++ b/SheetMetalTools.py
@@ -135,7 +135,8 @@ if isGuiLoaded():
                 the `FreeCADGui.Selection`.
 
             """
-            self.sync_selection()
+            _, obj_name, sub_name, _ = args
+            add_qtreewidget_item(self.widget, obj_name, sub_name)
 
         def clearSelection(self, *args):
             """Execute when clearing the selection.
--- a/SheetMetalTools.py
+++ b/SheetMetalTools.py
@@ -123,10 +123,7 @@ if isGuiLoaded():
             self.widget.clear()
 
             for obj in Gui.Selection.getCompleteSelection():
-                item = QtGui.QTreeWidgetItem(self.widget)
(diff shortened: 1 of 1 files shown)


shaise:
the addremove button is already connected here for all tasks.

image

cant you put your code there?

zokhasan:
You're right. So the change will be clearer. Thanks.

shaise:
Great feature!!
Thanks!

regarding delete button, I prefer the second one. (one delete button)

zokhasan:
PR version.

https://github.com/user-attachments/assets/c75c032e-0ff6-4b17-8a9b-c57787e0f2c6

As for the button, I tried to make several variants of it, but in the end I did not add it to this PR because at the moment I don't have a good implementation for it.

shaise:
Great!!! thanks!

PR #541, opened 2025-10-16
Allow formulars in cells of material definition sheets

UR-0:
Also distinguish between empty cells and cells containing a zero.


UR-0:
Current version fails importing attached material definition sheet, because of:
- column B contains formulars, such as "=0.65 + 0.5 * log10(A3)"
- column A contains zeros (e.g. R/t=0) for sharp bends
[Bend.FCStd.zip](https://github.com/user-attachments/files/22952228/Bend.FCStd.zip)
freecad-Screenshot-300

shaise:
Thanks! nice fix.

PR #546, opened 2025-12-16
unfolder: handle coaxial tangent cylinders

### LINKED ISSUE #545: Issue with Unfolding
Hello, attached is 1 zip file with 2 STEP files and 2 images in it. 

- "Sample_1.step" unfolds incorrectly, the flat doesn't show the flanges fully which have the hem bend, "Sample_1.jpg" shows the flanges which are not contained in the flat. What is the reason for this? 
- "Sample_2" unfolds incorrectly, the flat doesn't show the inside flanges, "Sample_2.jpg" shows the inside flanges which is not contained in the flat.  What is the reason for this? 

[Samples.zip](https://github.com/user-attachments/files/24165158/Samples.zip)

alexneufeld:
closes #545 

There are actually 2 possible tangency cases: 
image

### DIFF
--- a/SheetMetalNewUnfolder.py
+++ b/SheetMetalNewUnfolder.py
@@ -215,10 +215,17 @@ class TangentFaces:
         # approximately r1 + r2.
         return (
             SheetMetalTools.smIsParallel(c1.Axis, c2.Axis)
-            and abs(
-                c1.Center.distanceToLine(c2.Center, c2.Axis) - (c1.Radius + c2.Radius)
+            and (
+                abs(
+                    c1.Center.distanceToLine(c2.Center, c2.Axis)
+                    - (c1.Radius + c2.Radius)
+                )
+                < eps
             )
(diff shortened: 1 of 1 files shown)


shaise:
Fast response! thanks!

shaise:
@alexneufeld 
A non related question:
When the unfolder is executed internally all the flat faces are aligned to the global xy plane, and at the end, the sketches are left on the XY but the unfold solid is aligned back to the position of the selected face. Is there a reason for that other than compatibility with the older unfolder?

alexneufeld:
just compatibility. Solidworks and Catia do flat patterns 'in place', I think autodesk stuff (fusion, inventor), usually opens a separate tab when unfolding a part. I just copied what the old unfolder did. Feel free to change the new unfolder's behavior or make the in place unfolding optional.

shaise:
thanks!

PR #553, opened 2026-02-01
Update FreeCAD Forum announcement/discussion address

jblb:
to reflect the new domain freecad.org


shaise:
Thanks for the catch!

PR #562, opened 2026-03-09
Implement hem feature

### LINKED ISSUE #302: Feature: Simple way to make hems
Hems are often used in sheet metal. There are several kinds of hem (see picture below). Most of them can be done using `SMMakeWall` and maths to compute the right parameters. The rolled hem isn’t possible right now because it needs the leg length to be zero.

Commands for hem can inherit `SMMakeWall` and do some calculations to determine the bend radius and leg length. Here is a document with all the formulas: [sheetmetal hems.pdf](https://github.com/shaise/FreeCAD_SheetMetal/files/14463385/sheetmetal.hems.pdf). I have not described the rope hem because it needs two flanges so it is a more complex object.

image

pierreporte:
Resolves #302

This is my first code contribution to the workbench. I’m not a programmer, I just know a little bit of Python. I’ve never worked on an existing code base before.

What is needed to do, as I see it now:

- [x] When base feature is edited, the hem object breaks somehow
- [x] Icon
- [ ] Task panel
- [x] Add documentation for the math (basically it’s what I posted in #302)

What I don’t want to do with this PR:

* Improving support for miter
* Improving support for corner relief


pierreporte:
@shaise SheetMetalCmd.py is very big. Do you think that `SMWall` and `SMHem` classes (and accessories) should be in a separate file each?

maxwxyz:
@pierreporte 
SheetMetal_AddHem

pierreporte:
It’s exactly what I had in mind, thanks!

Though is this normal to have blue inside the bend?

image

maxwxyz:
ah no, will correct it

shaise:
@pierreporte 
Nice!!
Yes, please put smHem in a separate file.
Also, is there a task panel for that? 

shai

maxwxyz:
@pierreporte 
SheetMetal_AddHem

pierreporte:
Thanks @maxwxyz 

@shaise There is no task panel yet. I will certainly need help for it. Now I need to get the math right because I want to change the width users will enter, then I’ll split the file and try to solve issues when editing the tree. The task panel should be the last one.

shaise:
@pierreporte ,

No problem, do what you can, I will deal with the task panel.

pierreporte:
The new version uses the total width of the hem as input, including the bend part, not just the flat portion.

I added the open rolled hem and separated the open hem from the flat heam. I also added a degenerate case for the teardrop hem.

I tried to derive an explicit mathematical solution but it was too hard for me. Maxima and AI were able to do it for the closed teardrop hem, but it was a very big expression and I didn’t want to take the risk of writing it. For the open teardrop it would have been even more complicated. Thus I tried to find the solution numerically, considering that I knew a small interval where the solution is. I started with the bisection method, which has a guaranteed convergence, but is not very fast. Although it needs 29 iterations to reach $10^{-9}$ precision, the calculations lasts less than a millisecond. It is perfectly usable within FreeCAD. A document describing the math will be put in the repository after the merge.

The new mathematical expressions reuse the old one but I didn’t find the full derivation I did two years ago so it’s pretty light.

Now I need to test this thoroughly, and then we can do the task panel. In the meantime, where the checks should be implemented? In the actual feature code or in the task panel?

shaise:
hi @pierreporte ,

First of all: nice work! thanks.
Please update the disclaimer at the top of the file with the correct file name, and set your name at the copyright. (email is not necessary)
Before creating a task panel we need to discuss what properties we can remove from the original bend feature. (For example, I do not think we need extend1 and 2.
Also please explain what do you mean by checks? Do you mean to make sure the user input for the command is correct?
For selections, it is done before such that wrong selections will disable the icon.
For other things we need to take into account that the task panel is not available when freecad is in command line mode, so all essential code must be outside the gui system.

pierreporte:
I did what you asked for.

For the checks yes it is verifying that the input makes sense. I used the `raise` keyword to throw a `ValueError` exception with a message.

Now I’d like to hide or show some properties that I use for the hems depending on the type. For example I don’t need bend radius for flat hem or open hem, and I don’t need width for rolled hem. Do you have any idea how to do it? I’ve found a 10-year old forum thread that I have to read.

I have already removed some properties from the properties, and just remove `extend1` and `extend2`. There is no mitter for rolled hem but it exists for the other types so this is going to be conditional (like the previous paragraph).

shaise:
Hiding properties in the property table is a bit tricky as far as I remember, but I think this is less important since when we implement the task panel, we can handle hiding inputs there.
Let me know when you are more or less finished with the code and if you want me to implement the task panel.

pierreporte:
You are right that the task panel will hide the complexity of parameters but I think it would still be a good thing.

Besides that I think it's more or less ready. Let me just write how I would like the task panel to be organized so we have a starting point.

pierreporte:
With the last commit, the displayed properties change depending on the hem type.

One last thing: should I add a mode for the width type? Potential modes are: include bend (current), exclude bend.

Then it will be ready for review, and I’ll propose a task panel mock-up.

shaise:
> 
> One last thing: should I add a mode for the width type? Potential modes are: include bend (current), exclude bend.
> 
Do you mean something like the Length Spec of a standard bend? It can be nice but not critical IMHO.

pierreporte:
Yes exactly. It's easy to do so I'll do it.

pierreporte:
The new mode has been implemented. This week-end I make mock-ups for the task panel.

I noticed a problem: when testing with the 4th March weekly, the hem updates itself when parameters change without clicking recompute. With 18th March weekly, I need to manually recompute. Normal bends do update as expected. I’ve tested previous commits of this PR and they don’t recompute automatically with the new weekly. I haven’t investigated further.

pierreporte:
I specified the <kbd>Z</kbd> shortcut because it had to be something, but I didn’t think about it later. What should it be?

tigert:
image

option "opened" seems to work in reverse? otherwise seems to work nicely! this is sweet addition.

image

shaise:
> I specified the Z shortcut because it had to be something, but I didn’t think about it later. What should it be?

if h not available, z is fine.

pierreporte:
@shaise Here are task panel mockups. This is just the UI for hems. The tabs for things that weren’t deleted from the bend feature are still there (offsets and miter, except for rolled hem that has just offsets, and offset don’t have extend anymore).

For the default values that vary depending on hem type, I don’t know how it should be done.

Hems are closed by default.

Material position icons can be reused. For width mode we would need new ones.

---------------------------------------------

image

------------------------------

image

-----------------------------------

image

------------------------------------------

image

pierreporte:
@tigert I pushed the fix.

shaise:
@pierreporte ,
Standard bends also have perforations.
Do you think it is applicable for hems as well?
image

pierreporte:
I thought I removed them. No idea really, I felt like they were not relevant.

shaise:
> I thought I removed them. No idea really, I felt like they were not relevant.

Great, then. I feel the same.

pierreporte:
So I’m stopping here for the feature set, only bug fixes will be added in this PR.

Do you want to add the task panel to this PR or to do it independently?

shaise:
I plan to merge you code and then add the task panel.

shaise:
@pierreporte ,
Why do we have a separate type "open"? Why not use "Flat" type and Open checkbox like the others? it seems to me more consistent.

pierreporte:
It was the idea initially, but since they are referred to with different names are are considered two different types (or at least it is my understanding), it gelt a good idea.

shaise:
I thought about it and even if indeed it is a separate type, I think it is still more consistent to have it as a check box. So personally I prefer to change it, but I will accept any decision you make.

pierreporte:
I’ve looked around to see how the different programs answer this question. Programs that have one feature for flat and open hems: Catia, Inventor. Programs that have separate features: Solidworks, NX, Fusion 360, Solid Edge, ZW3D. Alibre doesn’t have a hem feature. It would be more logical to follow the majority.

shaise:
Great then! A separate feature it will be.
Working on the panel now.

shaise:
@pierreporte ,

Still working on the panel. After integration I want to remove some more unused properties (all those that uses offsets and reference angles)

Besides that there is a small issue with the properties hide/show feature. It seems this function produce errors when reloading a file:
```
18:46:59  pyException: Traceback (most recent call last):
  File "C:\Users\shai\AppData\Roaming\FreeCAD\v1-1\Mod\FreeCAD_SheetMetal\.\SheetMetalHem.py", line 361, in onChanged
    fp.setEditorMode("opening", hidden)
<class 'AttributeError'>: Property container has no property 'opening'
18:46:59  pyException: Traceback (most recent call last):
  File "C:\Users\shai\AppData\Roaming\FreeCAD\v1-1\Mod\FreeCAD_SheetMetal\.\SheetMetalHem.py", line 366, in onChanged
    fp.setEditorMode("rollangle", hidden)
<class 'AttributeError'>: Property container has no property 'rollangle'
```
I think that when you load the file, and freecad starts to populate the property window, it calls this function for each added prop, so this function try to access parameters that not loaded yet. It also means this function is called lots of times (I counted 128 for the attached file) 
We need to find a way that this function is called once per hem feature when loaded.
Attached ref file

[HemIcons.zip](https://github.com/user-attachments/files/26167764/HemIcons.zip)

pierreporte:
We could check if all the parameters that are used in `onChanged()` exist, like:

```python
def self onChanged(self, fp, prop):
    if exists(fp.HemType) and exists(fp.opened) and ...:
        if prop == "HemType":
            ...
    else:
        pass
```

but there may be a more clever way.

The actual check would be `hasattr(fp, "attribute name")`.

shaise:
I have set a new function:
```Python
    def setEditorMode(self, fp, prop, mode):
        if (hasattr(fp, prop)):
            fp.setEditorMode(prop, mode)
```
as a wrapper, and it seems to solve the problem. I still dont like it being called a gazilion times, but we can address it later
B.T.W dont make any changes to this branch until I commit my changes

shaise:
@pierreporte ,

I have merged your code and added my task panel. I did not bump the version yet. as
I'm working now to on the hide/view features in the task panel. Please make some tests an let me know if you find issues (aside from the hide/view fields)

shaise:
hide/show fields implemented.
let me know if you see any issues before i bump the version.

pierreporte:
It works great! There are just a few issues:

------------------

Double separator for flat hem:

image

--------------------

Issue with thickness outside position. I tested it at the time but it seems some cases are failing.

```
22:43:09  pyException: Traceback (most recent call last):
  File "/home/thomas/.local/share/FreeCAD/v1-2/Mod/sheetmetal/SheetMetalHem.py", line 390, in execute
    s, f = smBend(
           ^^^^^^^
  File "/home/thomas/.local/share/FreeCAD/v1-2/Mod/sheetmetal/SheetMetalCmd.py", line 1518, in smBend
    CutSolid = CutFace.extrude(FaceDir * -offset)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<class 'Part.OCCError'>: BRepSweep_Translation::Constructor
```

-------------------

Miter tab should be disabled for rolled hem.

shaise:
double separator fixed. miter hidden now. Version bumped

@pierreporte 
Thanks for your great work!!!!

PR #575, opened 2026-04-04
[BaseWall] Allow use of Draft Clone of Sketch (Part::Part2DObjectPython)

Syres916:
See https://forum.freecad.org/viewtopic.php?t=104406

@shaise I'm not aware of any issues allowing `Part::Part2DObjectPython`, by all means let me know if you've had issues in the past.

### DIFF
--- a/SheetMetalBaseCmd.py
+++ b/SheetMetalBaseCmd.py
@@ -269,6 +269,7 @@ if SheetMetalTools.isGuiLoaded():
             if not (selobj.isDerivedFrom("Sketcher::SketchObject")
                     or selobj.isDerivedFrom("PartDesign::ShapeBinder")
                     or selobj.isDerivedFrom("PartDesign::SubShapeBinder")
+                    or selobj.isDerivedFrom("Part::Part2DObjectPython")
                     ):
                 return False
             return True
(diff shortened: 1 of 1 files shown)


Roy-043:
The `Part::Part2DObjectPython` object _should_ only be used for 2D geometry. But up to FreeCAD v1.0.2 the Draft Workbench used it for [Draft Wire](https://wiki.freecad.org/Draft_Wire), [Draft BSpline](https://wiki.freecad.org/Draft_BSpline) and [Draft BezCurve](https://wiki.freecad.org/Draft_BezCurve) objects that can be 3D.

`Part::Part2DObjectPython` objects are therefore not guaranteed to be 2D. I suggest adding a test to verify this.

Syres916:
@Roy-043  thanks for the feedback, will improve the 2D detection

shaise:
@Syres916 ,
Thanks for the PR.
The base wall tool is for planar sketches but not only for closed ones. 
closed sketch:
image

open sketch:
image

is there a reason to limit it?

Roy-043:
In the Draft WB we use findPlane to check if an objects is planar:
`plane = shape.findPlane()`

This does not guarantee that the plane matches the XY-plane of the Placement of the object. I do not know if that is relevant here.

shaise:
it only needs to be planar, orientation is irrelevant.

Syres916:
@shaise having gone full circle, I've reverted the checks and will leave it for the existing code to raise the error if the user selects an invalid non-coplanar wire as it does currently.

shaise:
Great! sounds good to me

PR #595, opened 2026-05-19
Supports Simplified Chinese

HaydenHu:


shaise:
Please read here:
https://github.com/shaise/FreeCAD_SheetMetal/blob/master/Resources/translations/README.md
PR for updating translations is no longer accepted.
You must do it through https://crowdin.com/project/freecad-addons
There is already simplified Chinese translation there. please update and let me know when finished.

PR #598, opened 2026-06-05
Allow zero length initial perforation

rtm7777:
Sometimes for short faces I don't want to have perforated edges
This PR allows to set them to 0

image

### DIFF
--- a/SheetMetalCmd.py
+++ b/SheetMetalCmd.py
@@ -1815,7 +1815,7 @@ class SMBendWall:
                 FreeCAD.Qt.translate("App::Property", "Perforation Angle"),
                 0.0,
                 "ParametersPerforation")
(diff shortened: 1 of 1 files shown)


shaise:
Good idea. I will test it this weekend and merge.

shaise:
Hi,
It seems to work well.
However, why did you change the property from length to distance? is this necessary? Is some cases changing property type can break models that were saved in older versions of sheetmetal. (in this case it did not break because the properties are similar)

rtm7777:
Well, for some reason I thouhght I won't be able to set zero because length is positive number, but it works fine so I reverted it, good catch

shaise:
Great! Thanks for the contribution. merging.

Issue #460, opened 2025-02-01
Feature: Different relief shapes for corners (i.e. rounded relief shape for laser/plasma)


ma00aem:
### Describe the feature

Different relief shapes for corners. There is a rounded relief shape that is really good for laser/plasma cutting
search for :Laser weld relief shape corner

### Would you like to work on this issue?

None

luzpaz:
Here's a tip: if you make your proposal as elaborate as possible, there is a higher probability that it would be implemented. This Feature Request is very sparse.

shaise:
There is already a round relief shape to corners. can you attache images of the types you propose?

ma00aem:
Hi, excuse me for the lack of detail. I'm unsure if i can provide pictures from other softwares, so i tried to describe with words. Is it ok to make links to screenshots or oterwise i gladly can show you how it works in other software and give you a better understanding of what i try to describe.

shaise:
Link to screenshots is good enough

ma00aem:
https://help.autodesk.com/view/INVNTOR/2021/ITA/?guid=GUID-A7F8AB3C-A9AC-43AB-86FF-45133B124F7A
Se under Laser Weld Relief Shape

shaise:
Thank you, I understand now.

alexneufeld:
This documentation page has a better list of Inventors available relief shapes 
https://help.autodesk.com/view/INVNTOR/2021/ENU/?guid=GUID-EEFF7315-EF98-4922-AE22-09F9C6B3D488

You can also currently create the laser weld relief with a a slotted cutout on a solid body , then a 3d face offset 

![Image](https://github.com/user-attachments/assets/a093eee7-9cd9-48e5-a715-0176027392c4)

![Image](https://github.com/user-attachments/assets/e19ce171-1eb2-4482-93c8-6a759858bd3a)

[corner_relief.zip](https://github.com/user-attachments/files/18634470/corner_relief.zip)

Blendware:
Hi, if no one is currently working on this, I’d like to work on this feature.

shaise:
That will be great!

luzpaz:
@Blendware thank you. Hows the progress ?

Blendware:
@luzpaz  I think I should be done soon. I do push code regularly on my [fork](https://github.com/Blendware/FreeCAD_SheetMetal/tree/WeldCorner) if you'd like to take a look.

Currently I'm aware of 2 bugs:

[...]

PR #617, opened 2026-09-10
Adding feature for rounded relief corner #460

Blendware:
Relief size can be 0 and it will add a Shape with no radius:
image
Or it can have a radius by changing size property:
image

It needs at least 1 extended face  with an edge that is connected and parallel to the bend:
|Corner with extended face|After adding weld relief|
|-|-|
|image|image|


shaise:
when i try to change the corner from round to weld, I get the following error:
`" CornerRelief: At least 1 gap is needed"`

What does it mean?
see attached file.
[WeldCorner.zip](https://github.com/user-attachments/files/32145742/WeldCorner.zip)

Blendware:
Your file doesn't have a extended face at the corner:
image

Because there is no geometry to attach to it will fail
We might need to change the error message, I used "gap" because that's how the base shape type tub calls them.
The way how my algorithm works to find the length of the weld (4) is to check the connected edges from the corner 1
in sequence up to edge 4: 
image

I did think about making a way to attach gaps, but I wanted to keep that as a separated feature that needs its own discussion that I was planning to make after this one was merged.

shaise:
i still get an error even where closed gaps are enabled:
image

(same file, jut enable close gaps in base shape, using freecad 1.1)

Blendware:
Looks like removeSplitter() fails when size is > than intern radius of bend
Should I just disable it when it fails? should I move the radius of the corner towards inside?

shaise:
if the problem is only remove splitter, but the original solid was valid,
then yes, just ignore it. but usually remove splitter fail when the solid
is incorrect.

Blendware:
Looking further into it seems like it does give a boolean failure ~~on a older version of FreeCAD while checking but not on a newer version but removeSplitter() still fails even if geometry is valid~~.

shaise:
sometimes when boolean operations are done with overlapping vertex/edges it causes errors. I sometime add some epsilons to some dimensions to avoid that.

shaise:
Sorry to bother you again, but now circular and square corners fail (same file)
image

Blendware:
Are you using a absolute value that is too small? If not please provide the values you used.
if value is too small it would fail if when you switch because the square and circle where designed this way.

shaise:
Sorry, you are right, my mistake. Looks like everything is working now. Merging.
Thanks! this is a very useful addition.

PR #619, opened 2026-09-16
Fixing WeldCorner Bugs

Blendware:
- Adding Weld property to older versions when switching to Weld inside Task Gui
- Fixing Weld length on folds with 2 different angles
- Mitigate weld width returning None

There is still one known issue I've notices: when a wall is angled, the cutting operation can end up cutting the adjacent folds: 
<p float="left">
imageimage
</p>
That issue may take me several days to fix.


shaise:
Thanks for the fix!

